PR #5, opened 2019-09-24
Initial commit of reference implementation

tplooker:
This PR commits a reference implementation of VC-Authn using OpenID connect, whereby the OIDC interface leverages the IdentityServer C# implementation of an OpenID provider. Interactions involving verifiable credentials and DIDComm is achieved through controlling an instance of ACA-Py via its controllers REST interface.


tplooker:
@swcurran, fixed this with an interactive rebase.

.gitattributes
```diff
@@ -0,0 +1,10 @@
+# Set the default behavior, in case people don't have core.autocrlf set.
+* text=auto
+
+# Declare files that will always have LF line endings on checkout.
```
swcurran: Not sure that I like this in gitattributes. We have users on Mac, Linux and Windows. Pretty sure we leave this to be always auto.  OK for now, but will check with the team.

esune: I think this is the right thing to do - we do the same in other projects. This will effectively only be required in Windows systems for bash scripts, etc. as UNIX/Linux will already checkout using LF for line endings.

swcurran: Sounds good. As long as it's consistent with other projects.

Thanks for the follow up.

PR #22, opened 2019-12-19
fixed mismatch between oauth flow names and sequences

elf-pavlik:
It appears that sequence diagram had mismatch between oauth flow names and sequences. I only corrected them in diagram's source and before merging you may want to add commit to this PR with updated png rendering.


swcurran:
@tplooker - can you please confirm this change?  I don't know enough :-).  I'm happy to generate the image and replace it in the text.

@elf-pavlik - thanks for the update.

tplooker:
@swcurran @elf-pavlik, yes this is correct thanks for the change!

Issue #24, opened 2019-12-24
Controller Swagger UI not protected by API key


esune:
The swagger UI exposed by the vc-authn controller does not seem to be requiring the api key to be passed along in the requests to manage the presentation request configurations.

The functionality is already available in the code, so we need to check whether it has been misconfigured or it is broken and requires fixing.

PR #25, opened 2020-01-02
Fix controller API key header

esune:
This fixes the API key header setting being ignored, even when correctly configured, and makes the public endpoints protected.

Build and deployment configurations were also added: the files have been copied over as-is from the `law-society-demo`, omitting some of the default configurations that need to be tweaked for each deployment (e.g.: image namespace, route urls).


esune:
Fixes #24

PR #29, opened 2020-01-14
Demo apps for OIDC identity provider

esune:
The PR adds a demo application using `vc-authn-oidc` as Identity Provider directly, without the need of using an intermediary AIM solution.

The code has been updated to support the functionality, however further tweaks may/will be necessary to fully-implement all the standard claims for OIDC tokens (see [this](https://tools.ietf.org/html/rfc7519#section-4) for reference).


esune:
@swcurran, about the above points:
* we need `von-network`, or another ledger, to start `vc-authn-oidc`. I added a pre-requisite step that refers to the instructions to run `von-network`, similar to what we already do in the main project readme.
* replaced environment variables with the names you suggested, I agree it makes replacing them more straightforward.
* fixed typos (I hope all of them) and script permissions. Funny enough: my spell-checker wasn't highlighting "tunning" 😝 
* The reference to line 97 was pointing at the variables that would need to be used to connect to the database directly. I am now just referring to the `controller-db` configuration section and pointing to the file, but if it is confusing I could remove this altogether since the default configuration is shipped with the demo.
* still thinking about running everything from the same `manage` script: it is easy to compose docker-compose files, however I still need to duplicate environment variables and am wondering if I have a way of doing it without duplicating code. Automating the `ngrok` step will come next, that is more complex.

PR #59, opened 2020-04-14
Add support for custom label as referent in proof-request

esune:
Proof-Requests will now try to use the label defined in the configuration as referent, if available.
If the same label is used multiple times, the referent will try and disambiguate the repeated instances by adding a `~{counter}` suffix.
If no label is available, a GUID will be generated and used instead.

@andrewwhitehead I think using `~` as separator is fairly safe (maybe with the exception of very rare use cases), I just want to confirm it will not clash with anything in aca-py the way I am using it.
It doesn't appear to do so, I have used it on my machine withouth issues, but it doesn't hurt to have an educated opinion 🙂

### DIFF
--- a/oidc-controller/src/VCAuthn/Models/OptionalAttribute.cs
+++ b/oidc-controller/src/VCAuthn/Models/OptionalAttribute.cs
@@ -1,0 +1,11 @@
+using System;
+
+namespace VCAuthn.Models
+{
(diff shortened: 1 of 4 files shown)


andrewwhitehead:
I can't think of any special handling for a tilde there. Referents are treated as opaque strings.

esune:
Adding a tweak, please hold-off from merging for a minute

PR #90, opened 2020-09-26
make poll interval configurable

vitalrev:


oidc-controller/src/VCAuthn/IdentityServer/IdentityConstants.cs
```diff
@@ -46,6 +46,7 @@ public class IdentityConstants
         
         public const string ChallengeIdQueryParameterName = "pid";
         public const string ChallengePollUri = "vc/connect/poll";
+        public const int PollInterval = 2000;
```
esune, line 49: Any way we could make this configurable by reading an environment variable and fall back to the default value if that is not set?

I like the idea, however this requires the app to be rebuilt in order for the change to take effect.

esune (changes requested):

swcurran:
@esune -- given what we found with argparse in Python land, maybe there is the equivalent we could use here?

esune:
> 
> 
> @esune -- given what we found with argparse in Python land, maybe there is the equivalent we could use here?

Yes, it should be just a matter of binding the value to an environment variable or configuration rather than hard-coding it.

oidc-controller/src/VCAuthn/IdentityServer/IdentityConstants.cs
vitalrev: i've changed so that poll interval can be configured in appsettings.json and env var in manage script

esune (approved):

esune:
Thank you @vitalrev !

PR #91, opened 2020-09-27
Feature - oidc angular demo with keycloak and vc

vitalrev:


esune (changes requested):
Thanks for the submission, this looks good!

A couple comments:
- the new demo uses Angular 8: this is not a huge deal, however Angular * will be out of support in a couple of months. Is it possible to use Angular 9 or even 10 instead?
- the django demo connects directly to `vc-authn-oidc` without going through Keycloak. I think it might be better to do the same for the Angular otherwise we're demonstrating how to authenticate with Keycloak using a third-party IdP.

Let me know what you think! :)

vitalrev:
* yes, you are right... I try to update the angular version
* my first version of angular demo was exact the same (connect directly to controller without Keycloak). It worked but not so 100% clean, because the identity server (controller) doesn't implement the OIDC discovery document correctly.  
And, i would demonstrate how to integrate the verifiable auth into Keycloak OIDC as third-party IdP because I think the most people/projects know and use Keycloak as de facto standard. We can show how to add VC Auth as the alternative/additional authentication method... 
I mean, it can be a long process to switch the hole world from OIDC/OAuth to VC-Auth :-) 
Keycloak + VC-Auth as third-party can be the first step (hybrid) in that direction!

Let me know what you think!

esune:
Yes, you are correct when you say that the discovery document is not 100% complete/correct. This is because the VC provider is only capable of enabling the authentication flow, not authorization. It also does not hold any type of user information (the claims resented in the proof are just passed along in the id token upon successful authentication).

A also agree that using `vc-authn` through an Identity Broker such as Keycloak is likely going to the way to go at least initially, However I still think it probably is better to have a direct-connection demo rather than using Keycloak because:
- there is enough documentation on the web about how integrating a webapp with Keycloak works
- there are docs (just added recently!) in this repo that highlight how to configure a new client, and use Keycloak as example
- because of the fact that the discovery document is not complete, using Keycloak as demo may be misleading - meaning users may thing that `vc-authn` is providing all the token information the app is displaying, while it actually isn't and it is Keycloak that is doing the heavy lifting.

Lets ask @swcurran for an additional opinion.

swcurran:
I would rather have an Angular demo than not, so am happy either way, as long as the person coming into the repo gets the correct context in running the demo.  So having the demo as is, and even adding a "Help Wanted" issue to say it would be nice to have this not use Keycloak might be sufficient.

I'd definitely not want to leave this PR out indefinitely.

esune:
@vitalrev please let me know if you think you can update the Angular version, otherwise I will update the main readme and merge the code as-is.

vitalrev:
@esune sure, I will update the angular version. I hope next week...

vitalrev:
@esune I have updated angular version and other packages, please review again

esune (approved):

esune:
Thank you @vitalrev , merged!

PR #95, opened 2020-10-08
Fix broken authentication flow

esune:
Removed unused dependencies and services registered at startup.

Fixed broken authentication flow: the base url was being set incorrectly, resulting in a non-resolvable address in the QR code.

Tested in demo mode and it REALLY works this time 😅


docker/manage
```diff
@@ -178,7 +178,7 @@ configureEnvironment() {
   # controller
   export IDENTITY_SERVER_URL="${IDENTITY_SERVER_URL:-http://localhost:5000}"
   export IDENTITY_SERVER_API_KEY="controller-api-key"
-  export IDENTITY_SERVER_WEB_HOOK_URL=${IDENTITY_SERVER_URL}
+  export IDENTITY_SERVER_WEB_HOOK_URL=${IDENTITY_SERVER_WEB_HOOK_URL:-http://$DOCKERHOST:5000}
```
esune, line 181: This is the url used buy the agent to deliver webhooks, so it has to be on the internal docker network.

WadeBarnes (approved):

PR #107, opened 2020-11-24
Added BestPractices.md

esune:
Describe need to check which proof request configuration was responded to after login.


esune:
@swcurran please let me know what you think on this, the idea is to add any clarification/best practice that is required when using `vc-authn` for suthentication.

swcurran (approved):
lgtm

PR #116, opened 2021-01-04
ui tweaks to auth page

wadeking98:


wadeking98:
ui_changes

esune:
Added @swcurran to the review for a visual check.

What I would tweak:
- I would leave the "authentication required" header at the same level and color of the rest of the background
- I personally don't like having text/buttons above the QR code (I know this is what the mockup showed), and would personally like it better if the QR code and button were after the header (as before), and the descriptive text at the bottom.

Let's see what @swcurran thinks, I thin this is a mix of preferences and what I think would be more usable across desktop and mobile devices, but I am not a usability expert so...

wadeking98:
tweaked UI
qrchanges2

esune:
@swcurran can you please confirm this works for you and merge?

oidc-controller/src/VCAuthn/Views/Authorize/Authorize.cshtml
```diff
@@ -43,21 +43,29 @@
...
           </div>
 
+          <p class="info"><span class="info-header">MOBILE DEVICE USERS: </span> To authenticate, tap the button below</p>
           <b-button :href="btnUri" variant="outline-primary">
               <b-icon icon="box-arrow-up-right"></b-icon>
               Open in a Trusted Digital Wallet
           </b-button>
+
+          <p class="info"><span class="info-header">DESKTOP/LAPTOP USERS: </span>
+           To authenticate, present your digital credential by using your identity wallet phone app to scan the QR code below</p>
```
swcurran, line 64: The QR code is above now, right?  The "below" should read "above"?

swcurran (changes requested):

esune (approved):

PR #125, opened 2021-04-04
Clarify wallet network for email verification demo

### LINKED ISSUE #124: Clarify wallet network needed for email verification service in demo to prevent network mismatch error
While iterating through following demo instructions using the Trinsic Wallet:

https://github.com/bcgov/vc-authn-oidc/blob/master/docs/DemoInstructions.md

the "Get a verifiable email credential" section does not specify which **network** you should be using! For example, when I try the Sovrin production network - wallet default - I am able to get the connection added but not the pending credential offer! I get a network mismatch error. My assumption was to use BCovrin since its a BC related demo but that didn't work either.

iamgollum:
Resolves #124


esune (changes requested):
Please amend your commit to add DCO as it is required to merge the PR.

I would also tweak the documentation to require the agent to be on Sovrin Staging **before** trying to establish a connection with any of the demo issuers as it is more correct/straightforward from a workflow perspective (e.g.: fulfill all pre-requisites and then go get the credential).

iamgollum:
Thanks @esune for the quick turn around. I am with you on better placement for the workflow (honestly threw it without much thought as I had a hunch the community leaders would already have ideas they wanted there).

esune (approved):

PR #141, opened 2021-11-18
Align with OOB spec

amanji:
This PR introduces changes that align with the OOB specification and further enables support for connection-less presentation requests in mobile wallet implementations.

* OOB messages are base64 URL encoded as opposed to base64 encoded. This resolves issues with introducing special characters such as `+` in the encoded string which can carry meaning in a URL.
* Provides a mechanism to return decoded OOB messages in JSON format from the URL shortening service while still supporting redirect operations.

References:
1. https://datatracker.ietf.org/doc/html/rfc4648#section-5
2. https://github.com/hyperledger/aries-rfcs/tree/main/features/0434-outofband#url-shortening


swcurran (approved):
LGTM

PR #150, opened 2021-12-20
Update the demo to use port 5001 instead of 5000 to make it easier on a Mac

swcurran:
- Update all port 5000 references to 5001
- Updates to readme for demo from running locally


WadeBarnes:
I don't really like how the local docker environment's configuration is affecting the deployment configuration.  It should be possible to separate the two so the deployment configuration can be defined independently of the local docker configuration.  That would allow it to remain the same now, and change independently in the future if needed.

swcurran:
I think I know what you are saying (too much hardcoded?) but either way, it's a different change than this. All I did here was change the port 5000 references to 5001.

esune (changes requested):
I agree with @WadeBarnes: the changes should be restricted to the `docker` folder, otherwise they will cause a cascade of changes in repositories using this for their source (e.g.: [bcgov/a2a-trust-over-ip-configurations](https://github.com/bcgov/a2a-trust-over-ip-configurations)].

All you are looking for is to have it run on a different port locally, so that AirPlay will not cause a conflict (for now... :smirk:). Updating the code configuration and deployment configurations is therefore unnecessary.

swcurran:
Can you tell me the impacted files to undo?  Is it all of the yml files in the Openshift folder?

WadeBarnes:
All of the files in the `openshift` folder and perhaps `oidc-controller/src/VCAuthn/Properties/launchSettings.json` and `oidc-controller/src/VCAuthn/appsettings.json`.  You should be able to override the properties in the json files, if needed, using environment variables set in the `docker-compose.yml` file.  The syntax for overriding a .Net app setting with an environment variable is to use `__` to separate the json tree levels, so `UrlShortenerService.BaseUrl` becomes and environment variable named `UrlShortenerService__BaseUrl`.

esune:
> Can you tell me the impacted files to undo? Is it all of the yml files in the Openshift folder?

The only files being updated should be inside the `docker` folder. Everything else should not be modified.

swcurran:
Done, but wondering about the two files in `oidc-controller/src`? I think they are only for the demo, but wanted to confirm.

Thanks

swcurran:
That doesn't work -- checking why.  On running the demo, I got this error from ngrok:

`Failed to complete tunnel connection
The connection to https://8d4e-96-54-62-211.ngrok.io was successfully tunneled to your ngrok client, but the client failed to establish a connection to the local address localhost:5000.

Make sure that a web service is running on localhost:5000 and that it is a valid address.

The error encountered was: dial tcp [::1]:5000: connect: connection refused`

swcurran:
All good, I think -- I had reverted the ngrok config as well. Trying again.

swcurran:
OK - should be good to go.

docker/docker-compose.yml
```diff
@@ -111,7 +111,7 @@ services:
   controller-dev:
     image: vc-authn-oidc-controller-dev
     ports:
-      - "${IP:-0.0.0.0}:5000:5000"
```
esune, line 114: The right-hand port should be left as before, as it is internal to the Docker image and not mapped to the host.

docker/oidc-controller/Dockerfile
```diff
@@ -14,7 +14,7 @@ RUN dotnet publish -c Release -o /app/out --no-restore
 # Build runtime image
 FROM mcr.microsoft.com/dotnet/core/aspnet:3.1
 
-EXPOSE 5000
```
esune, line 17: Port should be left as `5000` as other services/configurations will depend on this when communicating on the internal Docker network.

docker/oidc-controller/Dockerfile.dev
```diff
@@ -13,4 +13,4 @@ ENV DOTNET_USE_POLLING_FILE_WATCHER 1
 
 WORKDIR /app
 
-ENTRYPOINT dotnet watch run --urls=http://+:5000
```
esune, line 16: Same as previous Dockerfile, leave as 5000 as se4rvices communicating on the internal Docker network will rely on this setting. The exposed port can/will be different (in docker-compose).

esune (changes requested):
A couple of port settings would be best reverted to limit the interactions with dependant projects.

esune:
> Done, but wondering about the two files in `oidc-controller/src`? I think they are only for the demo, but wanted to confirm.
> 
> Thanks

Those can actually be updated, as they are settings used by demo/local development and need to be mapped to the port bound on the host 👍🏻

There will also be projects that will require updating as the port used to communicate with `vc-authn` on localhost will have changed (e.g.: https://github.com/bcgov/vc-visual-verifier/blob/main/docker/manage#L225).

swcurran:
Updates made.  Started on the vc-visual-verifier changes, but the instructions are a bit lacking there.  Will open an issue on that repo.

esune:
> Updates made. Started on the vc-visual-verifier changes, but the instructions are a bit lacking there. Will open an issue on that repo.

Looks good to me now! I agree the visual verifier docs could be improved, it was put together a bit rushed and we never got a chance to go back and polish things there.

esune (approved):

WadeBarnes:
@swcurran, Another PR got merged before this one.  Could you please rebase these changes?

swcurran:
Done

PR #195, opened 2023-01-12
2.0 development

Jsyro:
Python (FastAPI, SQLModel, SQLAlchemy, Alembic) based solution using postgresql database.

TODO's

- [x] Add auto redirect on presentation verification. 
- [ ] No styling on Authorization screen
- [x] Injected 'Subject Identifier' into claims from verification_configuration. 
- [ ] ... probably a few more, will update when found.


oidc-controller/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM python:3.10
```
esune, line 1: Could we maintain the same pattern as before, and have a `docker` folder in the root with all the Docker resources (Dockerfiles, docker-compose, configurations mounted as volumes, etc.) and the helper `manage` script? This would be consistent with the old pattern used in the repo (making the transition simpler), as well as the pattern used for other DTS apps.

oidc-controller/api/authSessions/dependencies.py
```diff
@@ -0,0 +1,11 @@
+from fastapi import Depends
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from ..db.session import get_async_session
+from .crud import AuthSessionCRUD
+
+
+async def get_veriication_configs_crud(
```
esune, line 8: Nit.: typo `veriication`

oidc-controller/api/core/acapy/client.py
```diff
@@ -0,0 +1,133 @@
...
+        self.wallet_token = resp["token"]
+        logger.debug(f"<<< get_wallet_token")
+
+        return self.wallet_token
+
+    def create_presentation_request(
+        self, presentation_request_configuration: dict
+    ) -> CreatePresentationResponse:
+        logger.debug(f">>> create_presentation_request")
+        if not self.wallet_token:
```
esune, line 51: Note for the future: we will likely want to have the vc-authn agent run outside of an agency, with the agency agent being an option (but not mandatory). We can discuss further, but since the agent is read-only I feel like having a dedicated standalone service might be best.

Since this targets a tenanted wallet, we should add a `TODO` to handle different instances properly.

oidc-controller/api/core/aries/present_proof_presentation.py
```diff
@@ -0,0 +1,28 @@
...
+    id: str = Field(alias="@id")
+    type: str = Field(
+        "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/present-proof/1.0/request-presentation",
+        alias="@type",
+    )
+    request: List[PresentProofv10Attachment] = Field(
+        alias="request_presentations~attach"
+    )
+    comment: str = None
+    service: ServiceDecorator = Field(alias="~service")
```
esune, line 19: Let's add a `TODO` to switch to OOB protocol (this may not be the best spot though?).

oidc-controller/api/core/config.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+class EnvironmentEnum(str, Enum):
+    PRODUCTION = "production"
+    LOCAL = "local"
+
+
+class GlobalConfig(BaseSettings):
+    TITLE: str = "Traction"
+    DESCRIPTION: str = "A digital wallet solution for organizations"
```
esune, line 22: We may want to be able to override these at runtime

scripts/docker-compose.yaml
```diff
@@ -0,0 +1,100 @@
...
+      KEYCLOAK_LOGLEVEL: ${KEYCLOAK_LOGLEVEL}
+      ROOT_LOGLEVEL: ${KEYCLOAK_ROOT_LOGLEVEL}
+      KEYCLOAK_IMPORT: ${KEYCLOAK_IMPORT}
+    depends_on:
+      - keycloak-db
+    networks:
+      - vc_auth
+
+  keycloak-db:
+    image: registry.access.redhat.com/rhscl/postgresql-10-rhel7:latest
```
esune, line 72: We may want to upgrade to a newer version of postgres - I would go with the "vanilla", non RHEL version personally (we can also do this separately).

oidc-controller/api/authSessions/dependencies.py
Jsyro: fixed

oidc-controller/api/core/aries/present_proof_presentation.py
Jsyro: added todo to the `send_connectionless_proof_req` method that uses that class.

oidc-controller/api/core/config.py
Jsyro: fixed

scripts/docker-compose.yaml
Jsyro: quickly changing that to `postgres:12` cause the import to fail, will look into that in the future.

oidc-controller/Dockerfile
Jsyro: Got a `manage` file, and no .env file. Please take another look and let me know. Would be happy to change that in a follow up branch.

oidc-controller/api/core/acapy/client.py
esune: Will add a dedicated issue/task for this

PR #214, opened 2023-01-26
2.0: Move presentation claims out of top level of idtoken. 

Jsyro:
Breaking Change: 

All existing oidc configurations will need to be update to look for any imported attributes from the presentation under the `vc_proof_claims` member of the idtoken. 

Thoughts on the attribute name of `vc_proof_claims`.

### DIFF
--- a/oidc-controller/api/core/oidc/issue_token_service.py
+++ b/oidc-controller/api/core/oidc/issue_token_service.py
@@ -1,5 +1,7 @@
 import logging
+import json
 import uuid
+import dataclasses
 from datetime import datetime
 from typing import List, Dict, Any
 from pydantic import BaseModel
@@ -10,6 +12,7 @@ from ...verificationConfigs.models import VerificationConfig
 logger = logging.getLogger(__name__)
 
 
+@dataclasses.dataclass(frozen=True)
 class Claim(BaseModel):
     type: str
     value: str
@@ -20,7 +23,7 @@ class Token(BaseModel):
     issuer: str
     audiences: List[str]
     lifetime: int
-    claims: Dict[str, Claim]
+    claims: Dict[str, Any]
 
     @classmethod
     def get_claims(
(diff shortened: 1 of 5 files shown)


oidc-controller/api/core/oidc/issue_token_service.py
```diff
@@ -75,12 +77,13 @@ def get_claims(
...
-        claims.append(Claim(type="sub", value=sub_id_value))
-        claims += list(presentation_claims.values())
+        oidc_claims.append(Claim(type="sub", value=sub_id_value))
+        presentation_claims.values()
 
-        result = {c.type: c for c in claims}
-        logger.info("claims generated for the token:")
-        logger.info(result)
+        result = {c.type: c.value for c in oidc_claims}
+        result["vc_proof_claims"] = json.dumps(
```
esune, line 84: What about `vc_presented_attributes` as container name? This would be pretty self-explanatory and consistent with the Aries nomenclature as well. Thoughts?

esune (commented):
Could we make it so that presented attributes matching the name of [OIDC Standard Claims](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims) will be returned at the root level - populating the corresponding standard claim?

We could still return them as part of the presented attribute object, but this way it would make it seamless for OIDC clients to pick those values up without needing special mapping.

oidc-controller/api/core/oidc/issue_token_service.py
```diff
@@ -46,7 +47,9 @@ def get_claims(
             Claim(type="nonce", value=auth_session.request_parameters["nonce"])
         )
 
-        presentation_claims: Dict[str, Claim] = {}
+        presentation_claims: Dict[str, Claim] = {
+            "given_name": Claim("email", "email1@net.com")
```
esune, line 51: It looks like this made it's way through to the committed code

esune (changes requested):

oidc-controller/api/core/oidc/issue_token_service.py
Jsyro: Good catch, should be good to go now.

esune (approved):

PR #216, opened 2023-02-01
Use MongoDB instead of PostgreSQL

Jsyro:
We don't want to store the auth_sessions in persistent storage anyways. 
Verification configurations are lightweight.

We do not see a need for join operations or massive datasets.


oidc-controller/api/verificationConfigs/router.py
```diff
@@ -1,69 +1,65 @@
...
     VerificationConfigRead,
+    VerificationConfig,
 )
 
 router = APIRouter()
 
 
 @router.post(
-    "", response_model=VerificationConfigRead, status_code=http_status.HTTP_201_CREATED
+    "/",
```
esune, line 16: Status code seems to be missing, is this on purpose?

Jsyro: fixed, good catch!

esune (approved):

PR #219, opened 2023-02-07
Update manage script to automate set-up

### LINKED ISSUE #210: Update manage script to be more wizard-like
The `manage` script is meant to be a helper used to start the project, and should be as simple and automated as possible.

Some improvements to look into:
- [ ] Prompt to decide whether to use a multi-tenant agent, or the default dedicated instance
- [ ] If the multi-tenant configuration cannot be programmatically inferred, add prompts to collect it
- [ ] Execute `wallet_init.py` when necessary
- [ ] More? Assess what needed.

esune:
The `manage` script has updated to automate the steps to start-up `vc-authn` in various modes:
- the user will be prompted to select whether they want to run in single-tenant or multi-tenant/traction mode
- the user will be prompted to choose whether to use ngrok to expose the services or not

Both `curl` and `jq` are being executed in a docker container, removing the dependency for separate binaries on the path.

Ngrok containers have been moved to a separate compose project file: they need to be started before all of the other services in order to retrieve the public URLs and use set environment variables using their values.

I have not been able to test whether this works with `traction` or not as I cannot start the project in its current state - @Jsyro please let me know if you can run the code and it works.

Resolves #210


esune:
I am still trying to upgrade the `ngrok` containers to use the official, more regent image. Other than that the changes should be good to go.

docker/manage
```diff
@@ -201,33 +163,28 @@ configureEnvironment() {
...
+  export ST_ACAPY_ADMIN_API_KEY_NAME="x-api-key"
+
+  # agent
+  export AGENT_TENANT_MODE="${AGENT_TENANT_MODE:-single}"
+  export AGENT_HOST="http://aca-py"
+  export AGENT_NAME="VC-AuthN Agent"
+  export AGENT_HTTP_PORT=${AGENT_HTTP_PORT:-8030}
+  export AGENT_ADMIN_PORT=${AGENT_ADMIN_PORT:-"8077"}
+  export AGENT_ADMIN_URL=${AGENT_ADMIN_URL:-http://$AGENT_HOST:$AGENT_ADMIN_PORT}
+  export AGENT_ENDPOINT=${AGENT_ENDPOINT:-http://$AGENT_HOST:$AGENT_HTTP_PORT}
```
Jsyro, line 179: AGENT_HOST is set later in the options, so this gets prefixed with `http://aca-py` even when AGENT_HOST is set to `http://<traction-acapy-host>` later.

docker/manage
```diff
@@ -319,6 +255,69 @@ deleteVolumes() {
...
+        ;;
+      2)
+        read  -p "Please provide your tenant's Wallet ID:" MT_ACAPY_WALLET_ID
+        read  -p "Please provide your tenant's Wallet Key:" MT_ACAPY_WALLET_KEY
+        echo "AGENT_TENANT_MODE=multi" > .env
+        echo "MT_ACAPY_WALLET_ID=${MT_ACAPY_WALLET_ID}" >> .env
+        echo "MT_ACAPY_WALLET_KEY=${MT_ACAPY_WALLET_KEY}" >> .env
+        echo "ACAPY_HOST=http://ngrok-traction-agent" >> .env
+        echo "AGENT_ADMIN_PORT=8080" >> .env
+        break
```
Jsyro, line 277: Even using this option, the aca-py container is still started.

Jsyro (changes requested):
When using the 'multi-tenant' option, I ran into two issues. See comments.

docker/manage
esune: Tested locally, aca-py and wallet containers do not get started when selecting multi-tenant/traction option. Can you try pulling the code and trying again (after running `./manage rm`)? It might have been something I pushed after you checked out the code.

Jsyro: disregard, if the container exists, it will be restarted, if it is delete, ./manage correctly DOES NOT create the `aca-py` and `aca-py-ngrok` containers.

docker/manage
esune: The latest commit should fix this issue.

esune:
It looks like there is an issue with the newer `ngrok` images, I will wait until that is resolved before I attempt to upgrade again the configuration. See https://github.com/ngrok/docker-ngrok/issues/25

esune:
Added some tweaks that should make quality of life better when restarting the project while keeping the same settings, and refreshing ngrok endpoints if necessary. If this gets us close enough to where we want to be I'd lean towards merging/closing and logging more specific issues about what needs tweaking as we find out.

Jsyro (approved):
Great changes, still some minor exceptions I need for my development environment, but this should work for everyone else.

PR #221, opened 2023-02-10
In-demo instructions - tweaks - Security

Jsyro:
ngrok url for traction needs to be the tenant-proxy service
tokenurl for keycloak should work on import now. 

on screen instructions, most can be automated in future work

Basic secret header security to ver_config endpoints


oidc-controller/api/core/config.py
```diff
@@ -69,6 +67,8 @@ class GlobalConfig(BaseSettings):
     JWT_ALGORITHM = "HS256"
     JWT_ACCESS_TOKEN_EXPIRE_MINUTES = 300
 
+    CONTROLLER_API_KEY: str = os.environ.get("CONTROLLER_API_KEY", "change-me")
```
esune, line 70: Should we default to no api key if none is provided (e.g.: unprotected)?
It might be an edge case, but I think it makes sense.

esune (changes requested):

esune (approved):

PR #222, opened 2023-02-12
Remove Trinsic Wallet (not working) and added BC Wallet

swcurran:
Signed-off-by: Stephen Curran <swcurran@gmail.com>

Signed-off-by: Stephen Curran <swcurran@cloudcompass.ca>


Jsyro (approved):
We need to test and ensure compatibility with Lissi and Esatus, we will have to support them, but i've ran into issues. We hope a transition to OOB messages will improve compatibility.

swcurran:
Happy to remove others if they aren’t working. Better to push on them to improve, IMHO.

PR #230, opened 2023-02-23
Endpoint naming and provider config

Jsyro:
Add `/vc/connect` paths for backwards compatibility. Update provider config, still needs testing. 

Standard naming is found here: https://connect2id.com/products/server/docs/api. Some are standardized, some must be described in the `/.well-known/openid-configuration` response. 

PyOP library handles token generations and auth session tracking. some duplication should be trimmed, but the library now gives us the jwks endpoint and the wellknown config endpoint. as well client authentication.


docker/keycloak/config/keycloak_import.json
```diff
@@ -1181,7 +1181,7 @@
         "forwardParameters": "pres_req_conf_id",
         "tokenUrl": "http://host.docker.internal:5201/token",
         "authorizationUrl": "http://localhost:5201/authorize",
-        "clientAuthMethod": "client_secret_post",
+        "clientAuthMethod": "client_secret_basic",
```
esune, line 1184: Was this meant to be a change? I *think* the default behaviour is to send the secret as part of the `post`.

Either way, this is something we will need to flesh out as we work through the client authentication piece.

oidc-controller/api/core/oidc/provider.py
```diff
@@ -25,35 +25,51 @@
...
+
+
 provider = Provider(
-    signing_key, configuration_information, authz_state, client_db, Userinfo(user_db)
+    signing_key,
+    configuration_information,
+    AuthorizationState(subject_id_factory),
+    {"keycloak": kc_client},
+    Userinfo({"Jason": {"sub": "Jason"}}),
 )
```
esune, line 75: This piece will be configuration-driven in the future, correct?

docker/keycloak/config/keycloak_import.json
Jsyro: Yes, the library wanted it that certain way by default, but I'll take another look to see if the library's behaviour is configurable.

esune: Sounds good, will look at it as part of the next changes then.

oidc-controller/api/core/oidc/provider.py
Jsyro: yes, but I'm not sure about the user collection. It expects it to be a stored collection somewhere, so we may need a 'dummy' user record for the library... which is currently how this works.

Jsyro: yes, you are right. Not sure how to manage the 'User' collection, as it wants to be a dataset, but in our use case it is not, so we may need a single 'dummy record' for the library's sake.

esune (approved):

PR #237, opened 2023-03-10
Feature/unit testing

Jsyro:
First test. Had to change to Dependency Injection pattern for db connection as recommended by FastAPI docs.

Some explanation of env variables. 

tests can be run with `pytest` from top level folder.


docs/2_0_ConfigurationGuid.md
```diff
@@ -0,0 +1,5 @@
+### Environment Variables
```
esune, line 1: There's a typo in the filename 🙂 
I also think the 1.0 configuration guide still applies, possibly with minor tweaks, for what concerns integrating VC AuthN with Keycloak.

oidc-controller/api/authSessions/crud.py
```diff
@@ -11,26 +12,27 @@
...
 )
-from ..db.session import db, COLLECTION_NAMES
+from ..db.session import COLLECTION_NAMES
 
 
 logger = logging.getLogger(__name__)
 
 
 class AuthSessionCRUD:
-    @classmethod
```
esune, line 21: Is is expected for these annotations to disappear?

Jsyro: yes, the crud class is no longer a 'singleton', the db context gets passed on initialization of the instance. this makes it better and easier for testing databases to be injected instead of the runtime db.

docs/2_0_ConfigurationGuid.md
Jsyro: merged the two files and expanded on one more variable.

esune (approved):

PR #247, opened 2023-03-16
Publish image to GHCR

esune:
Build and push image to GHCR upon creation of a new release.

Target architectures are `linux/amd64` and `linux/arm64`: for some reason, build fails for `linux/386` when pulling `anyio` - will have to look at it in the future if we need `linux/386` images published.


Jsyro (approved):
Looks great!

PR #250, opened 2023-03-22
CRUD of OIDC clients

Jsyro:
Open to changing the model defaults, kind of a janky way to updating the client repository, hopefully it does not cause any issues later.

Stateless is basically what we are doing.
MongoWrapper is incompatible with our version of PyMongo, https://github.com/IdentityPython/pyop/issues/48
RedisWrapper would be a misuse of redis as we would need to add persistence to it.

### DIFF
--- a/oidc-controller/api/authSessions/crud.py
+++ b/oidc-controller/api/authSessions/crud.py
@@ -13,7 +13,6 @@ from .models import (
     AuthSessionPatch,
 )
 from api.db.session import COLLECTION_NAMES
-from api.core.oidc.provider import init_provider
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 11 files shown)


oidc-controller/api/authSessions/crud.py
```diff
@@ -43,19 +43,23 @@ async def get(self, auth_session_id: str) -> AuthSession:
 
         return AuthSession(**auth_sess)
 
-    async def patch(self, auth_session_id: str, data: AuthSessionPatch) -> AuthSession:
+    async def patch(self, id: str, data: AuthSessionPatch) -> AuthSession:
```
esune, line 46: Should we check whether the provided id is valid in other CRUD methods as well other that `post`?

esune (commented):
Looks good, only one small comment requiring feedback 🙂 
I am thinking we also may want to add capability to load a config file, but we can tackle that as a separate issue.

oidc-controller/api/authSessions/crud.py
Jsyro: Ok, i figured out why this is. 

The patch method called by the acapy-webhook handler passes in `id` as a bson.objectid which is a superclass of the PyObjectId that is being used.... but the get method is being passed `id` as a string, so i needed to convert it 

I can leave it and leave a comment as to why this is? or I can cast it to the str in the webhook handler and convert it to a PyObjectId in the CRUD method for consistency. 

I think i'd prefer the second. Thoughts?

esune: I agree with the second approach for consistency

PR #251, opened 2023-03-23
CRUD module tests for ver_configs and client_configs

Jsyro:
client configs need to be unique on `client_id` so it's now all referenced by that string member. 
had to tweak db_access for tests to work.


oidc-controller/test_signing_key.pem
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
esune, line 1: This is required to run tests: can we move it to the `test` folder instead?

oidc-controller/api/test_signing_key.pem
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
esune, line 1: Same as other `test_sigining_key.pem` file: if it is required for tests (and it looks like it may be the same file) we will want it in the `tests` folder. Even better: could the framework generate its own keyfile? If that is the case, these files should be gitignored.

oidc-controller/api/clientConfigurations/tests/test_cc_crud.py
```diff
@@ -0,0 +1,88 @@
...
+    redirect_uris=["http://redirecturi.com"],
+)
+
+
+@pytest.mark.asyncio
+async def test_client_config_get(db_client: Callable[[], MongoClient]):
+    client = db_client()
+    crud = ClientConfigurationCRUD(client.db)
+
+    client.db.get_collection(COLLECTION_NAMES.CLIENT_CONFiGURATIONS).insert_one(
```
esune, line 32: Nit.: typo `CONFiGURATIONS` - use all uppercase letters

oidc-controller/api/core/config.py
```diff
@@ -69,6 +69,7 @@ class GlobalConfig(BaseSettings):
     TRACTION_CORS_URLS: str = os.environ.get("TRACTION_CORS_URLS", "")
 
     # openssl rand -hex 32
+    SIGNING_KEY_FILENAME = os.environ.get("SIGNING_KEY_FILENAME", "signing_key.pem")
```
esune, line 72: If we can use this to load an externally-generated `pem` file we might want to rename to `SIGNING_KEY_FILE` and accept an absolute file path (I am thinking we might mount a volume in a deployment to preserve the keys between restarts - which path is that by default?).

oidc-controller/api/test_signing_key.pem
Jsyro: this is me in the middle of trying to fix an issue. sorry for the confusion, see above for the issue

oidc-controller/api/core/config.py
Jsyro: yea, this who thing needs a revisit. related to my comments above. I was just trying to get the tests work

esune (approved):

PR #256, opened 2023-04-03
Autogenerate signing_key.pem in docker/oidc-controller if required

### LINKED ISSUE #252: Autogenerate signing_key.pem file if not found.
nothing creates this required file. This shell command creates the file.

`openssl genrsa -out signing_key.pem 2048`

shaangill025:
- resolve #252

### DIFF
--- a/docker/manage
+++ b/docker/manage
@@ -140,6 +140,16 @@ build-oidc-controller-dev() {
     -f './oidc-controller/Dockerfile.dev' '..'
 }
 
+create-signing-key() {
+  #
+  # oidc-controller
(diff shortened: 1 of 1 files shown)


esune:
I changed the target branch - it needs to be based on and merged into the `2.0-development`.

This would work for local docker execution, but it won;t help for cloud deployments: what I think we should look at is having the oidc controller code generate the signing key file if it does not find one at the default location.

shaangill025:
Closing this PR as I was having issues merging changes, instead creating PR #260 which replaces this.

PR #263, opened 2023-05-04
Update Readme, tweaks to manage script

esune:
Updated readme to provide better instructions on how to run the project.
Updated `manage` script to exit with an error if agent/controller endpoints are not set correctly.
Removed dependency on `von-network`.


popkinj (approved):
This looks great! Thanks @esune .

Issue #271, opened 2023-05-18
Update Ngrok


popkinj:
Current version of Ngrok does not allow for an authenticated session. Requires the latest version to forward HTML.

Follow the example here.
https://github.com/bcgov/traction/pull/588/files

esune:
Resolved by #276

PR #273, opened 2023-05-18
Redirect 234

popkinj:
- When scanning a qr-code with the bc wallet, the default behaviour still occurs
- When scanning with a smartphones camera the user is redirected to a webpage
- The webpage is a configured variable

**Note:** Within the development environment, Ngrok is currently blocking the redirect to a webpage. This will get corrected by the [this ticket](https://github.com/bcgov/vc-authn-oidc/issues/271.)


oidc-controller/api/core/config.py
```diff
@@ -39,6 +39,8 @@ class GlobalConfig(BaseSettings):
     CONTROLLER_URL: str = os.environ.get("CONTROLLER_URL")
     # if ngrok is blocked by your local network, set this to your localhost for testing.
     CONTROLLER_URL_LOCAL: str = os.environ.get("CONTROLLER_URL_LOCAL", CONTROLLER_URL)
+    # Where to send users when trying to scan with their mobile camera (not a wallet)
+    CONTROLLER_CAMERA_REDIRECT_URL: str = os.environ.get("CONTROLLER_CAMERA_REDIRECT_URL")
```
esune, line 43: Could we provide a default, built-in page for users who don't have a separately-hosted page?

I like the fact any web-hosted page could be used, but I think it would be nice if we did not have to necessarily rely on that for the functionality.

docker/manage
```diff
@@ -166,6 +166,7 @@ configureEnvironment() {
     CONTROLLER_WEB_HOOK_URL="${CONTROLLER_WEB_HOOK_URL}#${CONTROLLER_API_KEY}"
   fi
   export ST_ACAPY_ADMIN_API_KEY_NAME="x-api-key"
+  export CONTROLLER_CAMERA_REDIRECT_URL="https://id.gov.bc.ca/static/selfsetup.html"
```
esune, line 169: Good find! I'll check-in with the BC Wallet team to see if they intend to update this page, or will be creating their own.

esune (changes requested):

oidc-controller/api/core/config.py
popkinj: I like this idea.
Are you thinking like a bare bones description page? Or should we do a little UI/UX exercise?

esune: I think, for now, we can have a pretty lightweight "it looks like you didn't use a wallet, try again" type of page. Styled so it is not terrible, but basic. We can do UI/UX as a second, separate step.

docker/manage
esune: FYI: https://github.com/bcgov/bc-wallet-mobile/issues/989#issuecomment-1567081034

popkinj:
Will continue next by filling the template with some minimal content and styling.

docker/manage
```diff
@@ -168,8 +168,8 @@ configureEnvironment() {
   export ST_ACAPY_ADMIN_API_KEY_NAME="x-api-key"
 
   # The redirect url can be a web link or the name of a template
-  export CONTROLLER_CAMERA_REDIRECT_URL="https://id.gov.bc.ca/static/selfsetup.html"
-  # export CONTROLLER_CAMERA_REDIRECT_URL="simple-wallet-instructions"
+  # export CONTROLLER_CAMERA_REDIRECT_URL="https://id.gov.bc.ca/static/selfsetup.html"
```
esune, line 171: If this is unnecessary (based on the next line), please remove.

oidc-controller/api/routers/presentation_request.py
```diff
@@ -27,9 +28,24 @@
...
-    """QR code that is generated should a url to this endpoint, which responds with the
-    specific payload for that given agent/wallet"""
-    logger.info("Scanning Application headers:: " + str(req.headers))
+    """
+      If the user scanes the QR code with a mobile camera,
+      they will be redirected to a help page.
+    """
+    # First prepare the response depending on the redirect url
+    if settings.CONTROLLER_CAMERA_REDIRECT_URL is None:
+        response = HTMLResponse("Please scan with a digital wallet")
```
esune, line 37: Could we just respond with the default `wallet_howto.html` if the envvar is not set to a custom value? This would simplify a bit the logic, since we ship the default file no matter what.

esune (commented):
Looking good! Just a couple comments/suggestions.

oidc-controller/api/routers/presentation_request.py
popkinj: Yeah sure... That makes more sense.

docker/manage
popkinj: Thanks for catching this @esune

popkinj:
Screenshot_20230530_114226_Chrome

esune (approved):
Looks great! Added a small quality-of-life change, other than that this is awesome :sligh

PR #276, opened 2023-05-24
Ngrok 271

popkinj:
Integration of Ngrok tokens so we can pass html to the user within the dev env.
Also updated the Ngrok image to the most recent version from the officially supported source.


docker/manage
```diff
@@ -330,6 +340,8 @@ function setNgrokEndpoints() {
     NGROK_AGENT_PORT=4059
   fi
 
+  sleep 60
```
esune, line 343: We should not have arbitrary sleep times in the script. What is this supposed to do?

esune (changes requested):

docker/manage
popkinj: It's the only thing allowing the build env to start properly.
I was thinking it's temporary. That was the new ticket I referred to in RC.

popkinj: Good eye though @esune

popkinj: If you want I can just comment it out for the PR... Then add a note beside it.

esune: I think we should figure out why this is needed, it should not be required to execute the script. DO you have any hints/ideas on why it is required? What happens when you remove it?

Issue #277, opened 2023-05-29
[2.0] Centralized logging


i5okie:
### Recommendation to implement centralized logging with [structlog](https://github.com/hynek/structlog)

AC

1) Logs should be structured, and streamed to stdout
2) Output should be determined based on the environment as per [Pretty Printing vs. Structured Output]
3) Format should be parameterized via an Environment Variable, defaulted to JSON
4) No `print` statements used to send output to console

PR #285, opened 2023-06-08
State 270

popkinj:
## Three States.
1. Not Scanned:
> Screenshot from 2023-06-08 10-57-24
2. Scanned and not yet approved:
> Screenshot from 2023-06-08 10-58-46
3. Scanned and approved: 
> Screenshot from 2023-06-08 10-56-36

**Possible Action Item**
Should the "Refresh QR-Code" button still be there? I think it was only relevant when we had a state of denial.... Which we don't.


docker/docker-compose.yaml
```diff
@@ -2,9 +2,9 @@ version: "3"
 services:
   controller:
     image: vc-authn-oidc-controller
-    depends_on:
-      controller-db:
-        condition: service_started
+    # depends_on:
+    #   controller-db:
+    #     condition: service_started
```
esune, line 7: Do you mind removing this completely? It appears that we keep commenting/uncommenting these lines 😅

oidc-controller/api/authSessions/crud.py
```diff
@@ -23,6 +23,7 @@ def __init__(self, db: Database):
         self._db = db
 
     async def create(self, auth_session: AuthSessionCreate) -> AuthSession:
+        print("auth_session in crud.py create: ", auth_session)
```
esune, line 26: Might want to remove this, unless we are using a logger. I added a comment to #277 to deal with other print statements as well (we should have non, logger should be used if meaningful).

oidc-controller/api/routers/acapy_handler.py
```diff
@@ -32,6 +32,7 @@ async def post_topic(request: Request, topic: str, db: Database = Depends(get_db
             auth_session: AuthSession = await AuthSessionCRUD(db).get_by_pres_exch_id(
                 webhook_body["presentation_exchange_id"]
             )
+            print("acapy_handler.py-webook:",webhook_body)
```
esune, line 35: Same as above, either use logger or remove print statement.

oidc-controller/api/templates/verified_credentials.html
```diff
@@ -217,45 +210,56 @@ <h1>Scan with a Digital Wallet</h1>
...
 
-          setTimeout(() => {
-            window.location.replace("{{callback_url}}", {method: 'POST'})
-          }, 3000);
+          // Turn off the spinner
+          qrcode.classList.remove("pending");
+
+          // setTimeout(() => {
+          //   window.location.replace("{{callback_url}}", {method: 'POST'})
+          // }, 3000);
```
esune, line 250: This needs to be uncommented, otherwise functionality will be broken (it will never redirect).

esune (changes requested):
There's a few things to tweak further, but it looks good otherwise 🙂 
I really like how the UI turned out, especially the animation after scanning the QR code.
The "refresh QR code" should be removed - it does not make sense to have it there anymore. We may re-introduce it once we are able to fully handle the success/failure states in verification.

esune:
@popkinj  there is an issue with the redirect that seems to have been caused by another commit (it will send you to a malformed URL). I have a change ready in another branch, disregard that when cleaning-up the requested changes 🙂

PR #286, opened 2023-06-09
Project cleanup, handle correctly all verification states

esune:
Changes include:
- general project cleanup
- fixed a bug that prevented a successful redirect to be completed due to a malformed URL
- introduced extra validation states for an authorization request, in order to handle unknown/pending/success/failure states
- tweaked the proof validation to pick-up the `verified` attribute rather than the `state` as the former indicates whether the proof passed, while state just that it was completed


popkinj (approved):
Looks good.

popkinj (approved):
Thanks for the fix @esune

PR #290, opened 2023-06-21
Proof Request Timeout 207

popkinj:
The front end polls the api for Proof Request status. The currently supported types are:
  - not_started
  - pending
  - verified
  - failed
  - **expired**

This PR is for the later status of `expired`.
The status is updated on the API side during initial request and in turn with each poll. I believe it would be a good improvement in the future to have the back end update the status on a timed process or queue without interaction with the front end.

Screenshot from 2023-06-21 11-47-08

Currently the proof request timeout is set to 5 seconds. This value can be adjusted as an environment variable [see here](https://github.com/bcgov/vc-authn-oidc/blob/timeout-207/docker/manage#L172)


oidc-controller/api/routers/oidc.py
```diff
@@ -37,6 +39,18 @@ async def poll_pres_exch_complete(pid: str, db: Database = Depends(get_db)):
...
+    """
+     Check if proof is expired. But only if the proof has not been started.
+     NOTE: This should eventually be moved to a background task.
+    """
+    if auth_session.proof_expired_time < datetime.now() and auth_session.proof_status == AuthSessionState.NOT_STARTED:
+        logger.info("PROOF EXPIRED")
+        auth_session.proof_status = AuthSessionState.EXPIRED
+        await AuthSessionCRUD(db).patch(
+            str(auth_session.id), AuthSessionPatch(**auth_session.dict())
+        )
```
esune, line 52: This looks extremely similar to the code in `acapy_handler.py`: any chance we can consolidate it/not have this duplication? If the duplication is needed, could you elaborate on why?

oidc-controller/api/authSessions/models.py
```diff
@@ -28,6 +28,7 @@ class Config:
 
 class AuthSession(AuthSessionBase, UUIDModel):
     proof_status: AuthSessionState = Field(default=AuthSessionState.NOT_STARTED)
+    proof_expired_time: Union[datetime, None] = Field(default=None)
```
esune, line 31: You could likely reuse `expired_timestamp` (even renaming it) in `AuthSessionBase`: I believe it was introduced for this purpose, but never wired-up (I might be wrong, so double checking is best).

oidc-controller/api/authSessions/models.py
```diff
@@ -41,4 +42,5 @@ class AuthSessionCreate(AuthSessionBase):
 
 class AuthSessionPatch(AuthSessionBase):
     proof_status: AuthSessionState = Field(default=AuthSessionState.PENDING)
+    proof_expired_time: Union[datetime, None] = Field(default=None)
```
esune, line 45: Same as previous comment

esune (changes requested):

oidc-controller/api/routers/oidc.py
popkinj: Yeah... I saw this possible duplication earlier. I just didn't know what sort of voodoo magic the older logic was doing. Thought it was some kind of keycloak thing so I didn't want to touch it.
If you're sure this is a duplication, I can consolidate.

oidc-controller/api/authSessions/models.py
popkinj: Same here. Didn't know what `expired_timestamp` was for.

oidc-controller/api/authSessions/models.py
popkinj: Same as above... Can consolidate if it won't break anything.

oidc-controller/api/authSessions/models.py
esune: I'd use that instead of adding a new property. It should be fairly quick to check whether it is being used somewhere or just set and never used afterwards.

oidc-controller/api/routers/oidc.py
esune: It looks like the code is basically doing teh same steps - at least in part. I'd assess whether it makes sense to consolidate rather than have them duplicated, that's all

popkinj: I'm on the same page. I'd rather not have duplication.

amanji (approved):

PR #297, opened 2023-07-24
Move logging to structlog and centralize logging 

### LINKED ISSUE #277: [2.0] Centralized logging

Gavinok:
This PR resolves #277 by integrating structlog into the logging system. A new config format using json is added to allow both the default logger to be adjusted as needed as well as additional config related to structlog. Both pretty printing and json logging can now be toggled via an environment variable `LOG_WITH_JSON` as well as a new `logconfig.json` file setting in case no envonrment variable is provided. If neither of these are set then we fall back to json logging as requested. in addition all print statments have been removed. 

Due to the limitations of structlog and the limited ability to modify 3rd party code structlog's processors where used to override the logging handled by uvicorn manually. `logconfig.json` can be adjusted to wrap other loggers from other libraries as needed for users. 

- [X] Logs should be structured, and streamed to stdout
- [X] Output should be determined based on the environment as per [Pretty Printing vs. Structured Output]
- [X] Format should be parameterized via an Environment Variable, defaulted to JSON
- [X] No print statements used to send output to console

In addition more work is necessary to format database logging from the *-db services. The latest acapy version seems to support `--log-json-fmt` which should be enough to help us support this in the `vc-authn-aca-py` service.

Example json logging
image

Example pretty printing
image


oidc-controller/api/core/config.py
```diff
@@ -1,12 +1,98 @@
...
+    level=logging.DEBUG,
+)
+
+# Use environment variable to determin logging format
+# fallback to logconf.json
+# finally default to true
+use_json_logs: bool
+
+time_stamp_format = "%Y-%m-%d %H:%M.%S"
+with open((Path(__file__).parent.parent / "logconf.json").resolve()) as user_file:
```
esune, line 27: It would be nice to allow the path to the configuration file to be overridden, and use this default file if nothing is specified.

The main reason is that `logconf.json` is placed with the source code and mounting a file/folder at this location might result complicated, while referencing a file located elsewhere in the filesystem will be easier for users to manage.

oidc-controller/api/core/config.py
```diff
@@ -1,12 +1,98 @@
...
+    stream=sys.stdout,
+    level=logging.DEBUG,
+)
+
+# Use environment variable to determin logging format
+# fallback to logconf.json
+# finally default to true
+use_json_logs: bool
+
+time_stamp_format = "%Y-%m-%d %H:%M.%S"
```
esune, line 26: We should allow users to override this, and fall back on ISO timestamp by default (based on info [here](https://www.structlog.org/en/stable/standard-library.html) the default format would be `iso`).

oidc-controller/api/core/config.py
```diff
@@ -1,12 +1,98 @@
...
+    structlog.processors.TimeStamper(fmt=time_stamp_format),
+]
+
+renderer = (
+    structlog.processors.JSONRenderer()
+    if use_json_logs
+    else structlog.dev.ConsoleRenderer()
+)
+
+# override uvicron logging to use logstruct
```
esune, line 55: Nit.: typo `uvicorn`

oidc-controller/api/core/config.py
```diff
@@ -1,12 +1,98 @@
...
-logger = logging.getLogger(__name__)
+from pathlib import Path
+
+logging.basicConfig(
+    format="%(message)s",
+    stream=sys.stdout,
+    level=logging.DEBUG,
+)
+
+# Use environment variable to determin logging format
```
esune, line 21: Nit.: typo `determine`

oidc-controller/api/core/config.py
```diff
@@ -1,12 +1,98 @@
...
+
+for _log in ["uvicorn", "uvicorn.error"]:
+    # Clear the log handlers for uvicorn loggers, and enable propagation
+    # so the messages are caught by our root logger and formatted correctly
+    # by structlog
+    logging.getLogger(_log).handlers.clear()
+    logging.getLogger(_log).addHandler(handler)
+    logging.getLogger(_log).propagate = False
+
+# This is already handled by our middlewear
```
esune, line 79: Nit.: `middleware`

oidc-controller/api/core/config.py
```diff
@@ -48,7 +134,9 @@ class GlobalConfig(BaseSettings):
 
     ACAPY_AGENT_URL: Optional[str] = os.environ.get("ACAPY_AGENT_URL")
     if not ACAPY_AGENT_URL:
-        print("WARNING: ACAPY_AGENT_URL was not provided, agent will not be accessible")
+        logger.info(
+            "WARNING: ACAPY_AGENT_URL was not provided, agent will not be accessible"
+        )
```
esune, line 139: Might want to set this log level to `warning, based on the message contents.

oidc-controller/api/main.py
```diff
@@ -64,12 +70,38 @@ def get_application() -> FastAPI:
     )
 
 
+# TODO clean up try catch
```
esune, line 73: Same as above

oidc-controller/api/main.py
```diff
@@ -64,12 +70,38 @@ def get_application() -> FastAPI:
...
+    return response
+
+
 @app.on_event("startup")
 async def on_tenant_startup():
     """Register any events we need to respond to."""
     await init_db()
     await init_provider(await get_db())
-    logger.warning(">>> Starting up app ...")
+    logger.warning(">>> Starting up app new ...")
```
esune, line 104: This could be better logged as `info`. Also not sure about "new`1 at the end.

oidc-controller/api/main.py
```diff
@@ -15,12 +21,12 @@
 
 from api.core.oidc.provider import init_provider
 
+logger: structlog.typing.FilteringBoundLogger = structlog.getLogger(__name__)
+
 # setup loggers
 # TODO: set config via env parameters...
```
esune, line 27: Is this work still in TODO, or should this be removed?

oidc-controller/api/main.py
```diff
@@ -15,12 +21,12 @@
...
 logging_file_path = (Path(__file__).parent / "logging.conf").resolve()
-logging.config.fileConfig(logging_file_path, disable_existing_loggers=False)
-
-logger = logging.getLogger(__name__)
+# structlog.config.fileConfig(logging_file_path, disable_existing_loggers=False)
```
esune, line 29: Please remove if unnecessary

oidc-controller/api/routers/presentation_request.py
```diff
@@ -104,5 +104,5 @@ async def send_connectionless_proof_req(
             service=s_d,
         )
         msg_contents = msg
-    print(msg_contents.dict(by_alias=True))
+    logger.info(msg_contents.dict(by_alias=True))
```
esune, line 107: Wondering if this should be a `debug` statement to avoid over-printing to logs when not necessary.

esune (changes requested):

esune:
Looking good overall! Just a few suggestions and nitpicky comments to keep the code as clean as possible 🙂

oidc-controller/api/core/config.py
i5okie: Setting the logconf.json file should probably be exposed via ENVVAR, and as @esune noted, default to `logconf.json`

oidc-controller/api/core/config.py
i5okie: Same here, these settings should be exposed via ENVVARs and defaulted as mentioned above.

oidc-controller/api/core/config.py
i5okie: Logging level setting should be exposed via an ENVVAR as well.

esune (approved):
👍🏻

PR #303, opened 2023-07-31
Update demo to run on M1 systems

esune:
Updated vue app Dockerfile to run on M1 Mac architecture, added retry handling fro ngrok tunnel detection

### DIFF
--- a/docker/manage
+++ b/docker/manage
@@ -367,8 +367,6 @@ function getNgrokUrl() {
   _output_file=$2
   _target_variable=$3
 
-  sleep 40
-
   docker run --rm curlimages/curl -L -s $_url > $_output_file
 
   NGROK_URL=$(docker run --rm -i ghcr.io/jqlang/jq:1.7rc1 < $_output_file --raw-output '.tunnels | map(select(.name=="command_line")) | .[0] | .public_url')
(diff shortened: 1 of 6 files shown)


esune:
I think I got it sorted out now. @popkinj mind testing on your machine if the `null` controller url is already an issue?
@WadeBarnes let me know if you see anything odd that I might have overlooked.

swcurran:
Does this mean we now have a working M1?

docker/manage
```diff
@@ -346,27 +352,50 @@ function setNgrokEndpoints() {
...
+    else
+      return 0
+    fi
+  }
+
+  local startTime=${SECONDS}
+  while ! extractUrl; do
+    printf "."
+    local duration=$(($SECONDS - $startTime))
+    if (( ${duration} >= 60 )); then
```
WadeBarnes, line 385: I would make the timeout configurable with a 60 second default.

WadeBarnes (changes requested):

docker/manage
esune: I thought about that, it seemed though like 60s would be the worst case scenario and in most cases it should really work after a 5-6 seconds so adding another configuration setting might be overkill?

esune:
> Does this mean we now have a working M1?

Yep, it should work on M1 without issues now - unless I missed something of course 🤣

WadeBarnes (approved):

popkinj (approved):
This looks like it'll do the trick.

PR #307, opened 2023-08-04
Websocket 305 - Fixing websocket with https.

popkinj:
A small update to the websocket URL.

Previously the websocket only worked in development environments (http). This will hopefully allow traffic over https as well.


swcurran (approved):
Progress!  I can now get the QR code scanned. New issues following that - I’ll add another ticket or two.

Issue #300, opened 2023-07-25
Update chart readme


esune:
Acceptance Criteria 

Update chart readme to be Bitnami-style, as already completed for Traction (see https://github.com/bcgov/traction/blob/2c6e437cddf1b4aac24ac1e0af12ba3cdbbdc6fe/charts/traction/README.md)

swcurran:
I'm not sure it is a good idea to document all of the ACA-Py startup parameters in each Helm chart, as in the example linked in the Acceptance Criteria.  Any chance there is a good way to document them/enable their use once in ACA-Py, and then reference where ever they are needed -- Helm charts, docker compose startups, etc.  The startup configs in ACA-Py are adjusted regularly, and we need a way to manage the (way too many!) options. Or at least, when we update a Helm Chart or docker compose configuration to use a new version, we pull in the new version of ACA-Py configuration settings.

Likewise, we need a way to document document the configuration parameter/environment variables that are used on top of ACA-Py and in a deployment.  Particularly important are the ports, and the common parameters that are shared across components -- e.g. the configuration parameters for a database that also have to be known to the components that are clients of the database.

It would be nice to have a clear view of where the list of settings can be found, and where overrides can be set, for the two deployment models we have -- Helm Charts and docker compose.

esune:
The Helm chart documentation will be only for the services specifically maintained in the repository - in this case the `vc-authn` controller. We will not be documenting all ACA-Py options here, but it will likely be a task for when we push an "official" ACA-Py chart to be used as a dependency (in which case the chart would live in ACA-Py's repo).

swcurran:
I was more talking about the example linked in the acceptance criteria.  The Traction Helm Chart looks like it has all of the ACA-Py config points documented.

esune:
Ah, gotcha. In the case of Traction we will have to see how to decouple the ACA-Py chart - I am not yet sure about it. The issue in that case is that Traction is an ACA-Py plugin and therefore the cart needs to run a modified version of ACA-Py. It should be easier for VC-AuthN since it's a "plain" agent running in read-only mode.

Issue #296, opened 2023-07-18
Add ACA-Py as chart component/dependency


esune:
Currently, an instance of ACA-Py (and corresponding wallet) is required to be deployed and running *BEFORE* the `vc-authn` chart can be installed in a k8s namespace.

Acceptance Criteria:
- Integrate ACA-Py as a sub-chart for vc-authn, as already implemented with Traction (see https://github.com/bcgov/traction/pull/672 ).

PR #315, opened 2023-08-14
[Helm Chart] Add ACA-Py as chart component/dependency

i5okie:
## Add ACA-Py as a Chart component
  - Add ACA-Py as a Chart component configurable through updated Values file
  - Document Values file and create a Readme
  - Add Postgres chart dependency for agent database
  
#### Addresses the following issues:
- https://github.com/bcgov/vc-authn-oidc/issues/296
- https://github.com/bcgov/vc-authn-oidc/issues/300


charts/vc-authn-oidc/templates/agent/deployment.yaml
```diff
@@ -0,0 +1,131 @@
...
+          imagePullPolicy: {{ .Values.acapy.image.pullPolicy }}
+          command:
+          - /bin/bash
+          args: [
+           "-c",
+           "aca-py start \
+           --inbound-transport http '0.0.0.0' {{ .Values.acapy.service.httpPort }} \
+           --outbound-transport http \
+           --admin '0.0.0.0' {{ .Values.acapy.service.adminPort }} \
+           --endpoint https://{{ include "vc-authn-oidc.host" . }} \
```
esune, line 46: We should likely use the `ACAPY_ENDPOINT` envvar/setting (see https://github.com/hyperledger/aries-cloudagent-python/blob/cbbcc8df2dbf1314e313adc170d3a0f77cbfed9c/aries_cloudagent/config/argparse.py#L606) and not force the protocol to `https`, but inherit what the route is using (unlikely to not use SSL, but having a hard-coded value here seems like it could potentially cause issues).

charts/vc-authn-oidc/templates/deployment.yaml
```diff
@@ -54,36 +55,46 @@ spec:
...
+            - name: CONTROLLER_CAMERA_REDIRECT_URL
+              value: {{ .Values.controllerCameraRedirectUrl }}
+            - name: CONTROLLER_PRESENTATION_EXPIRE_TIME
+              value: {{ .Values.controllerPresentationExpireTime | quote }}
             - name: ACAPY_AGENT_URL
-              value: {{ .Values.acaPy.agentUrl }}
+              value: https://{{ include "acapy.host" . }}
             - name: ACAPY_ADMIN_URL
-              value: {{ .Values.acaPy.adminUrl }}
+              value: https://{{ include "acapy.admin.host" . }}
```
esune, line 70: Same comment as before about hard-coding http scheme for URLs.

esune (changes requested):

charts/vc-authn-oidc/README.md
```diff
@@ -0,0 +1,242 @@
...
+
+## Installing the Chart
+
+To install the chart with the release name `my-release`:
+
+```console
+git clone https://github.com/bcgov/vc-authn-oidc.git
+cd vc-authn-oidc
+helm install my-release charts/vc-authn-oidc
+```
```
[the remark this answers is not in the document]
i5okie: updated.

charts/vc-authn-oidc/README.md
```diff
@@ -0,0 +1,242 @@
...
+| `acapy.argfile.yml.auto-respond-messages`         | Automatically respond to basic messages indicating the message was received. Default: false.                                                                                                                                                                                                                                            | `true`                                 |
+| `acapy.argfile.yml.auto-verify-presentation`      | Automatically verify a presentation when it is received. Default: false.                                                                                                                                                                                                                                                                | `true`                                 |
```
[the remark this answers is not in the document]
i5okie: The `argfile.yml` is dynamically built from the values. A couple of the values are also used in the helm chart logic.
I am leaning towards leaving it until we publish a GA ACA-Py chart.

charts/vc-authn-oidc/templates/agent/deployment.yaml
i5okie: Dynamically determining the protocol used in the endpoint is non-trivial. There are a number of caveats, is the end user using OpenShift Route object, Enables the Ingress object within the Values, or manually creates an Ingress object?

Perhaps we could add `agent.secureEnpoint` parameter in the Values, to specify wether to use `http` or `https`?

charts/vc-authn-oidc/templates/deployment.yaml
i5okie: Is it safe to assume that if the agent endpoint is using `https://` then the admin url is also using `https://`?

charts/vc-authn-oidc/templates/agent/deployment.yaml
esune: If the consumer is creating endpoints manually, they will have to tweak things anyway so I am not too concerned about that. I would, however, want things to work out-of-the-box when deploying the chart as-is. My main gripe is that if they are NOT using TLS when deploying the solution (unlikely in OpenShift as it is provided by the cluster, more likely for oher k8s instances used for prototyping) there will be a scheme mismatch between `http` ingresses and `https` protocol required in the chart.

I am actually not sure the agent would work without SSL enabled on the routes, so maybe all we need to do is put proper instructions/warnings?

charts/vc-authn-oidc/templates/deployment.yaml
esune: I would assume both endpoints will be using the same http scheme

i5okie: Updated `ACAPY_ADMIN_URL` to use internal admin url which points to agent's service and admin port.

charts/vc-authn-oidc/templates/agent/deployment.yaml
```diff
@@ -0,0 +1,131 @@
...
+          imagePullPolicy: {{ .Values.acapy.image.pullPolicy }}
+          command:
+          - /bin/bash
+          args: [
+           "-c",
+           "aca-py start \
+           --inbound-transport http '0.0.0.0' {{ .Values.acapy.service.httpPort }} \
+           --outbound-transport http \
+           --admin '0.0.0.0' {{ .Values.acapy.service.adminPort }} \
+           --endpoint {{ include "acapy.agent.url" . }} \
```
esune, line 46: Could use the `ACAPY_ENDPOINT` environment variable.

esune (changes requested):

charts/vc-authn-oidc/templates/agent/deployment.yaml
i5okie: updated.

esune (approved):
Looks good to me, let's test it out in the wild 👍🏻

PR #320, opened 2023-08-29
Deprecate name in favour of names array in requested_attributes

esune:
The `names` array is more flexible as it allows proofs to define one or more attributes to be requested from the same credential.
Instead of trying to handle exclusiveness between `name` and `names`, this change deprecates the single-attribute restriction in favour of the array that can, if needed, be used with one element definition only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ curl -X 'POST' \
     "requested_attributes": [
 
       {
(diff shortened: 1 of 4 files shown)


Jsyro (approved):
I LIKE THIS!! YAY

PR #338, opened 2023-09-18
Feature/332 fix middleware [2.0]

jamshale:
When the logging middleware function called `call_next` and it experienced an exception then the response object is not available and would throw a response not available in locals exception that would hide the real exception. It is un-catchable in the middleware.

This checks if response object exists in the locals in the finally block, and if it does then logs the message, and if it doesn't then it assumes it's an exception and looks in traceback,  logs the stack trace and returns a 500 response with a more informative message.

The json logger didn't work with the stacktrace from call_next, which would mean the production stack traces would be invisible. Now this explicitly logs the stacktrace from the middleware if `LOG_WITH_JSON` is True.

The stacktrace for the json format (production) is a bit ugly but I can't think of a way to really make them nice so I just log it as one big string.

### DIFF
--- a/oidc-controller/api/core/logger_util.py
+++ b/oidc-controller/api/core/logger_util.py
@@ -1,8 +1,9 @@
 import structlog
 import time
-
-
 from typing import Callable, Any
+
+logger = structlog.getLogger(__name__)
+
 
 
 def log_debug(func: Callable[..., Any]) -> Callable[..., Any]:
@@ -20,3 +21,11 @@ def log_debug(func: Callable[..., Any]) -> Callable[..., Any]:
         return ret_val
 
     return wrapper
+
+
+def extract_error_msg_from_traceback_exc(format_exc) -> str:
+    try:
+        return format_exc.splitlines()[-1].split(": ")[1:][0]
+    except Exception:
+        logger.error(f"Failed to extract error message from traceback: {format_exc}")
+        return "Unknown error"
--- a/oidc-controller/api/core/logger_util.py
+++ b/oidc-controller/api/core/logger_util.py
@@ -2,9 +2,6 @@ import structlog
 import time
 from typing import Callable, Any
 
-logger = structlog.getLogger(__name__)
-
-
 
 def log_debug(func: Callable[..., Any]) -> Callable[..., Any]:
     def wrapper(*args, **kwargs):
--- a/oidc-controller/api/core/logger_util.py
+++ b/oidc-controller/api/core/logger_util.py
@@ -21,11 +21,3 @@ def log_debug(func: Callable[..., Any]) -> Callable[..., Any]:
         return ret_val
 
     return wrapper
-
-
-def extract_detail_from_exception(exception_only_list) -> str:
-    try:
-        return exception_only_list[0].split(": ")[1:][0].rstrip()
-    except Exception:
-        logger.error(f"Failed to get exception details from: {exception_only_list}")
-        return "Unknown error"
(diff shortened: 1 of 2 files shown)


oidc-controller/api/main.py
```diff
@@ -90,14 +90,24 @@ async def logging_middleware(request: Request, call_next) -> Response:
...
+            logger.info("processed a request",status_code=response.status_code,process_time=process_time)
+        # Otherwise, extract the exception from traceback
+        else:
+            logger.info("failed to process a request", status_code=500, process_time=process_time)
+
+            if os.environ.get("LOG_WITH_JSON") is True:
+                logger.error(traceback.format_exc())
+
+            base_error_message = f"Failed to execute: {request.method}: {request.url}"
+            extracted_error = traceback.format_exc().splitlines()[-1].split(": ")[1:][0]
```
jamshale, line 107: This needs to be refined. Doubt it would work for all errors.

oidc-controller/api/main.py
```diff
@@ -77,9 +79,7 @@ def get_application() -> FastAPI:
 
 @app.middleware("http")
 async def logging_middleware(request: Request, call_next) -> Response:
-    # clear the threadlocal context
     structlog.threadlocal.clear_threadlocal()
-    # bind threadlocal
```
jamshale, line 82: I felt like these were redundant.

oidc-controller/api/core/logger_util.py
```diff
@@ -20,3 +21,11 @@ def wrapper(*args, **kwargs):
         return ret_val
 
     return wrapper
+
+
+def extract_error_msg_from_traceback_exc(format_exc) -> str:
+    try:
+        return format_exc.splitlines()[-1].split(": ")[1:][0]
```
jamshale, line 28: I'm not completely sure how resilient this will be. It gets the last line of the exception and then gets the message from after the colon.

jamshale:
Ugh. I messed up a rebase.

oidc-controller/api/core/logger_util.py
esune: I am not a big fan of having ad-hoc text handling to parse this. Did you look at - potentially - leveraging something like https://docs.python.org/3/library/traceback.html ?

jamshale: Ya. I'm not a fan of it either. I was trying different methods from this library and this was the best I had come up with. 

There is a library logging to the console with out the apps logger with logs like `Unknown client_id 'keycloak'`. You can see it if you don't push the initial configs. I'm not really sure why this is happening. 

I was thinking about looking into that library and see how it's parsing the exception and why it's logging to console.

esune: That would be https://github.com/IdentityPython/pyop when checking whether an OIDC client is registered with VC-AuthN. Could be a good idea, but they might just be catching a specific error/exception.

jamshale: Ya that wasn't helpful. I changed it to be a bit cleaner but it's still doing some text parsing. 

Another option is just not returning any details for 500 responses and developers would need to look in the logs, but having some information like this could be useful to end users. Then if there was a 500 error we could see the stack trace in the logs. Catch the exception correctly and raise an accurate HTTPException with details.

jamshale: I think this is the way to go. Making the changes.

esune (approved):
👍🏻

Issue #334, opened 2023-09-14
Add unit testing framework for critical code sections [2.0]


jamshale:
Unit testing as a whole is lacking but there is some critical logic areas that should be improved. Especially with the handling of complex objects. The unit testing framework including fixtures and mocks should be expanded and critical functions (mostly in the core) should be tested to make sure they are handling data correctly and throwing expected errors.

esune:
Resolved by #339

PR #339, opened 2023-09-19
#334 Add unit tests for critical code sections [2.0]

jamshale:
I'm sure more testing and better testing can be done but I think this is good enough for this task.
Only testing on python 3.11 now. The project is not compatible with 3.10.

I think the best testing is done in `test_issue_token_service.py` . It takes expected presentation payloads and assure they generate claims for valid inputs. I think this could be expanded further but it will likely come in handy as examples of how to send different payloads through the tests and will hopefully find any errors with expected functionality if a developer makes changes in the `get_claims` function.

Added some mock payloads which can be altered to do accurate testing and for context.

There is other testing in the acapy directory which is only somewhat useful. Still think it's nice to have for extra documentation and as a foundation for future features and proper testing.

Added a few TODO comments.

Here is the current coverage:

image

### DIFF
--- a/oidc-controller/api/core/acapy/tests/test_client.py
+++ b/oidc-controller/api/core/acapy/tests/test_client.py
@@ -1,0 +1,216 @@
+import json
+
+import mock
+import pytest
+from api.core.acapy.client import (CREATE_PRESENTATION_REQUEST_URL,
+                                   PRESENT_PROOF_RECORDS,
+                                   PUBLIC_WALLET_DID_URI, WALLET_DID_URI,
+                                   AcapyClient)
+from api.core.acapy.config import MultiTenantAcapy, SingleTenantAcapy
+from api.core.acapy.models import CreatePresentationResponse, WalletDid
+from api.core.acapy.tests.__mocks__ import (create_presentation_response_http,
+                                            presentation_request_configuration)
+from api.core.config import settings
+
+
+@pytest.mark.asyncio
+@mock.patch.object(settings, "ACAPY_TENANCY", None)
+async def test_init_no_setting_returns_client_with_single_tenancy_config():
+    client = AcapyClient()
+    assert client is not None
+    assert isinstance(client.agent_config, SingleTenantAcapy) is True
+
+
+@pytest.mark.asyncio
+@mock.patch.object(settings, "ACAPY_TENANCY", "single")
+async def test_init_single_returns_client_with_single_tenancy_config():
+    client = AcapyClient()
+    assert client is not None
+    assert isinstance(client.agent_config, SingleTenantAcapy) is True
+
+
+@pytest.mark.asyncio
+@mock.patch.object(settings, "ACAPY_TENANCY", "multi")
+async def test_init_multi_returns_client_with_multi_tenancy_config():
+    client = AcapyClient()
+    assert client is not None
+    assert isinstance(client.agent_config, MultiTenantAcapy) is True
+
+
+@pytest.mark.asyncio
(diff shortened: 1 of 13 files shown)


jamshale:
Only done one critical function so far but opening for visibility.

oidc-controller/api/core/oidc/issue_token_service.py
```diff
@@ -32,7 +32,7 @@ class Token(BaseModel):
 
     @classmethod
     def get_claims(
-        cls, pres_exch: Dict, auth_session: AuthSession, ver_config: VerificationConfig
```
jamshale, line 35: This parameter is no longer used. The auth session model generates it as a property.

oidc-controller/api/routers/oidc.py
```diff
@@ -165,8 +165,7 @@ async def post_token(request: Request, db: Database = Depends(get_db)):
 
     auth_session = await AuthSessionCRUD(db).get_by_pyop_auth_code(model.get("code"))
     ver_config = await VerificationConfigCRUD(db).get(auth_session.ver_config_id)
-    presentation = client.get_presentation_request(auth_session.pres_exch_id)
-    claims = Token.get_claims(presentation, auth_session, ver_config)
+    claims = Token.get_claims(auth_session, ver_config)
```
jamshale, line 168: The AuthSession model does this now.

oidc-controller/api/core/acapy/tests/test_client.py
```diff
@@ -0,0 +1,216 @@
...
+        status_code=200,
+    )
+
+    with mock.patch.object(CreatePresentationResponse, "parse_obj", return_value={'result': 'success'}):
+        client = AcapyClient()
+        client.agent_config.get_headers = mock.MagicMock(
+            return_value={'x-api-key': ''})
+        presentation_request = client.create_presentation_request(
+            presentation_request_configuration)
+        assert presentation_request is not None
```
esune, line 55: Would it make sense to check some basic fields in the resulting presentation request, to ensure it matches expectations?

oidc-controller/api/core/acapy/tests/test_client.py
```diff
@@ -0,0 +1,216 @@
...
+    with mock.patch.object(CreatePresentationResponse, "parse_obj", return_value={'result': 'success'}):
+        client = AcapyClient()
+        client.agent_config.get_headers = mock.MagicMock(
+            return_value={'x-api-key': ''})
+        try:
+            presentation_request = client.create_presentation_request(
+                presentation_request_configuration)
+            assert presentation_request is not None
+        except AssertionError as e:
+            assert e is not None
```
esune, line 76: Is this exception meant to be "anything", or are we expecting a specific exception to be raised here? If we're expecting a specific exception, could we assert on type and/or message to check it's what we are expecting and not something else?

esune (commented):
Excellent work! Added a couple of notes that can likely be extended to the remaining, similar scenarios.

Regarding the assertion to check for JSON objects: we are definitely expecting to receive valid JSON,if we use `json.loads()` to deserialize the payload it might be unnecessary since an exception would/should be thrown there if the object is not valid JSON?

oidc-controller/api/core/acapy/tests/test_client.py
jamshale: I think that would be a good idea, yes. More info in my other comment. Currently is just checking for 200 response and response isn't empty. Other exceptions are unhandled.

esune (approved):

PR #349, opened 2023-09-22
Add backwards compatibility setting for token claims

### LINKED ISSUE #347: Add backwards-compatibility setting for token claims
A major shift from VC-AuthN 1.0 is the introduction of the `vc_presented_attributes` to wrap the presented attributes into one object and avoid the need to know ahead of time the claim names used for a proof-request and create named mappers.

As a migration step from 1.0 to 2.0, however, we want to introduce a compatibility setting that will still allow named mappers to work and allow a transition period for client applications to adapt to the new pattern.

Token generation is handled in the [get_claims}(https://github.com/bcgov/vc-authn-oidc/blob/1e3ec9234d9ebb5f9c4201fb7d40d96a264b04fd/oidc-controller/api/core/oidc/issue_token_service.py#L34) method. It should be updated to add ALL presented attributes as named claims at the root of the token.

Suggesting using `USE_V1_COMPATIBILITY` as flag to be set to `true`/`false`.

Examples for a proof request resulting in `city` and `province` being shared, the expected token should look as follows:
- when  `USE_V1_COMPATIBILITY=false`
```json
{
  ...
  "vc_presented_attributes": {
    "city": "Victoria",
    "province": "BC"
  }
}
```
- when `USE_V1_COMPATIBILITY=true` (note the extra token claims)
```json
{
  ...
  "vc_presented_attributes": {
    "city": "Victoria",
    "province": "BC"
  },
  "city": "Victoria",
  "province": "BC"
}
```

jamshale:
Add the environment variable and then add the attributes a separate key/value pairs to the claim object outside of the `vc_presented_attributes` object.

Set the ver_config with optional include_v1_attributes field to default False. Then when getting the claims it decides to add the attributes individually based of this ver_config setting. 

This is the claim object pulled from logs when running demo.

> {
      'pres_req_conf_id': 'verified-email', 
      'acr': 'vc_authn', 
      'nonce': 'Y-JG5tXSWN1nplex-k653A', 
      'sub': 'test@email.com', 
      'vc_presented_attributes': '{"email": "test@email.com"}', 
      'email': 'test@email.com'
 }

Resolves #347

### DIFF
--- a/oidc-controller/api/core/config.py
+++ b/oidc-controller/api/core/config.py
@@ -184,7 +184,6 @@ class GlobalConfig(BaseSettings):
         os.environ.get("USE_OOB_LOCAL_DID_SERVICE", False)
     )
     SET_NON_REVOKED: bool = bool(os.environ.get("SET_NON_REVOKED", True))
-    USE_V1_COMPATIBILITY: bool = bool(os.environ.get("USE_V1_COMPATIBILITY", False))
     
     class Config:
         case_sensitive = True
--- a/oidc-controller/api/core/config.py
+++ b/oidc-controller/api/core/config.py
@@ -184,7 +184,8 @@ class GlobalConfig(BaseSettings):
         os.environ.get("USE_OOB_LOCAL_DID_SERVICE", False)
     )
     SET_NON_REVOKED: bool = bool(os.environ.get("SET_NON_REVOKED", True))
-
+    USE_V1_COMPATIBILITY: bool = bool(os.environ.get("USE_V1_COMPATIBILITY", False))
+    
     class Config:
         case_sensitive = True
 
--- a/oidc-controller/api/core/oidc/issue_token_service.py
+++ b/oidc-controller/api/core/oidc/issue_token_service.py
@@ -11,7 +11,6 @@ from pydantic import BaseModel
 from ...authSessions.models import AuthSession
 from ...verificationConfigs.models import ReqAttr, VerificationConfig
 from ..models import RevealedAttribute
-from ..config import settings
 
 logger = structlog.getLogger(__name__)
 
@@ -111,7 +110,7 @@ class Token(BaseModel):
 
         # TODO: Remove after full transistion to v2.0
         # Add the presentation claims to the result as keys for backwards compatibility [v1.0]
-        if settings.USE_V1_COMPATIBILITY:
+        if ver_config.include_v1_attributes:
             for key, value in presentation_claims.items():
                 result[key] = value.value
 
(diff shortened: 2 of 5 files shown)


oidc-controller/api/core/config.py
```diff
@@ -184,7 +184,8 @@ class GlobalConfig(BaseSettings):
         os.environ.get("USE_OOB_LOCAL_DID_SERVICE", False)
     )
     SET_NON_REVOKED: bool = bool(os.environ.get("SET_NON_REVOKED", True))
-
+    USE_V1_COMPATIBILITY: bool = bool(os.environ.get("USE_V1_COMPATIBILITY", False))
```
jamshale, line 187: Not sure if default should be true?

oidc-controller/api/core/oidc/issue_token_service.py
```diff
@@ -107,6 +108,13 @@ def get_claims(
         result[PROOF_CLAIMS_ATTRIBUTE_NAME] = json.dumps(
             {c.type: c.value for c in presentation_claims.values()}
         )
+
+        # TODO: Remove after full transistion to v2.0
+        # Add the presentation claims to the result as keys for backwards compatibility [v1.0]
+        if settings.USE_V1_COMPATIBILITY:
```
jamshale, line 114: I'm definitely not a python expert. Probably a slicker way to do this for loop.

oidc-controller/api/core/config.py
esune: I think `False` is fine: this is a migration step we will want to get away from whenever possible.

oidc-controller/api/verificationConfigs/models.py
```diff
@@ -41,6 +41,7 @@ class VerificationProofRequest(BaseModel):
 class VerificationConfigBase(BaseModel):
     subject_identifier: str = Field()
     proof_request: VerificationProofRequest = Field()
+    include_v1_attributes: Optional[bool] = Field(default=False)
```
jamshale, line 44: This doesn't necessarily need to be optional. Thought it made sense this way as it is being removed eventually.

esune: I agree with the approach 👍🏻

esune (approved):

esune:
@jamshale before merging, just confirming you tested having an "older" release with stored configs and upgrading and it works as expected?

jamshale:
@esune I haven't ran any testing where I started a v1.0 and upgraded to v2.0. I can do that.

jamshale:
@esune Ok I tested this now and it works as expected. 
The ver_config after upgrade doesn't have the `include_v1_attributes` field because it is optional, and it works as before. But then you can PATCH the field either as true or false and get the expected values in the attribute payload.

PR #364, opened 2023-10-12
Ensure user authentication sessions are independent

### LINKED ISSUE #362: Authorization sessions are being polluted by attempts on a separate proof
## Scenario:
Two proof requests are configured:
- `verified-email`, which does not use a subject identifier from the proof
- `accredited-lawyer-bcpc`, which uses a subject identifier from a claim in the proof request
Authentication requests are submitted by two separate clients (Keycloak Realms), one using `verified-email` and the other `accredited-lawyer-bcpc`.

## Behaviour:
1. Authenticate with VC-AuthN using the client requesting `verified-email` proof, everything seems to be working fine: users can authenticate and their subject id is generated by the service.
2. Authenticate with VC-AuthN using the client requesting the `accredited-lawyer-bcpc` proof configuration: user seems to sign-in successfully, using the assigned claim as subject id.
3. Authenticate again using the client requesting the `verified-email` proof: the subject identifier for the user matches the value generated in the previous request that used the `accredited-lawyer-bcpc` proof-configuration.

A guess is that `pyop` is not generating subject ids properly and this is causing the user database to only ever have *one* user that gets concurrently accessed by multiple sessions - which is very bad.

esune:
Resolves #362

Previously, the user authentication sessions was always leaning on the same "fake user" in the `Userinfo` database to store data temporarily. This lead to issues when multiple concurrent authentication requests were attempted: data from a user authentication session would pollute other user authentication sessions, leaking information.

The fix involved:
- defining a custom `VCUserinfo` class that overrides the methods in `Userinfo` responsible for returning data
- ensuring each authentication request is associated with a new, random user identifier to prevent polluting existing data
- in order to support both scenarios were the proof defines/doesn't define a `subject_identifier`, the generation of the `claims` object was updated to only include the `sub` claim if the proof-request defines one, and ensuring the token endpoint accounts for the custom sub, if present.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,10 +122,10 @@ After all these steps have been completed, you should be able to authenticate wi
 
 To connect a debugger to the `vc-authn` controller service, start the project using `DEBUGGER=true ./manage start` and then launch the debugger, it should connect automatically to the container.
 
-This is a sample debugger launch configuration for VSCode that can be used by adding it to `launch.json`:
+This is a sample debugger launch configuration for VSCode that can be used by adding it to `launch.json`, it assumes a `.venv` folder containing the virtual environment was created in the repository root:
 ```json
 {
-    "version": "0.1.0",
+    "version": "0.1.1",
     "configurations": [
         {
             "name": "Python: Debug VC-AuthN Controller",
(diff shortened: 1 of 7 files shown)


jamshale (approved):
I'd need to study this more to fully understand why the problem occurred.

It looks like you know what you're doing though :+1:

esune:
> I'd need to study this more to fully understand why the problem occurred.
> 
> It looks like you know what you're doing though 👍

Added two notes to try and clarify, but I am confident this is the right approach. It definitely was painful to figure out what the underlying library was doing 😅

PR #377, opened 2023-10-27
Properly toggle JSON logging

### LINKED ISSUE #363: JSON logger cannot be disabled
The json logger controlled by the `LOG_WITH_JSON` environment variable does not seem to be disabled when turning the flag to `false`. Additionally, doing so will cause stack traces to not be logged in the console.

Acceptance criteria:
- JSON logging can be turned on/off with the environment variable
- logging works correctly both when using JSON and standard printing to console

Gavinok:
This PR resolves #363
Since there are multiple ways individuals could pass a true or false as an environmental variable I added a dedicated function strtobool based on [distutils.util.strtobool](https://docs.python.org/3.9/distutils/apiref.html#distutils.util.strtobool) which will be removed in python 3.12. In addition, I have made the logging of stack traces independent of logging format.

Here is the regular text logging:

Here is the json version:

### DIFF
--- a/oidc-controller/api/main.py
+++ b/oidc-controller/api/main.py
@@ -92,15 +92,22 @@ async def logging_middleware(request: Request, call_next) -> Response:
     finally:
         process_time = time.time() - start_time
         # If we have a response object, log the details
-        if 'response' in locals():
-            logger.info("processed a request", status_code=response.status_code, process_time=process_time)
+        if "response" in locals():
+            logger.info(
+                "processed a request",
+                status_code=response.status_code,
+                process_time=process_time,
+            )
         # Otherwise, extract the exception from traceback, log and return a 500 response
         else:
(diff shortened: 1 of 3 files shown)


jamshale (approved):
Looks good to me. I would consider using snake_case for the function name `str_to_bool` . Makes it a bit more readable and clear imo.

oidc-controller/api/main.py
```diff
@@ -92,15 +92,22 @@ async def logging_middleware(request: Request, call_next) -> Response:
     finally:
         process_time = time.time() - start_time
         # If we have a response object, log the details
-        if 'response' in locals():
-            logger.info("processed a request", status_code=response.status_code, process_time=process_time)
+        if "response" in locals():
```
jamshale, line 95: You don't have to change this but I think in the future I would handle this by declaring response as `None` and then checking if `response == None` instead of checking the variable string name is in the locals array.

esune (approved):
Looks good to me 👍🏻

PR #376, opened 2023-11-20
Provide unique, consistent user identifier for repeated authentication attempts

Gavinok:
When authenticating with VC-AuthN, there are currently two options for generating a `subject_identifier` for the authenticated user:
- Using one of the proof-request attributes, if available and configured in the proof-configuration
- Generating a new identifier for each authentication attempt

The first option is only available IF the VCs used to authenticate provide a unique identifier: this would be configured in the proof-configuration to be used as `sub`, and would be passed along to the relying party that initiated the authentication request.

The issue with this pattern when integrating VC-AuthN into an AIM system such as Keycloak is that, unless a unique user identifier is generated by VC-AuthN, it will not be possible to associate roles to the user account in the AIM system as every authentication attempt will appear as a new user trying to access the system.

**Proposed solution:**
Rather than generating a new random user identifier for every authentication attempt, VC-AuthN would generate a consistent user identifier by hashing a canonicalized version of the JSON containing the claims obtained from the proof-request.

Possible library to canonicalize JSON: https://pypi.org/project/canonicaljson/

The underlying Pyop library will generate a new value for the `sub` if non is provided. A couple of places we could look at plugging-in the generation of the subject identifier value:
- [here](https://github.com/bcgov/vc-authn-oidc/blob/0cc400380541e7fcf9d2fb75f909782380984150/oidc-controller/api/routers/oidc.py#L173-L184) to generate the sub *BEFORE* pyop takes over to generate the token
- by implementing a new [SubjectidentifierFactory](src/pyop/subject_identifier.py) matching our criteria and injecting it in the code at the right time ([here](https://github.com/bcgov/vc-authn-oidc/blob/0cc400380541e7fcf9d2fb75f909782380984150/oidc-controller/api/core/oidc/provider.py#L118)).

### DIFF
--- a/oidc-controller/api/core/oidc/issue_token_service.py
+++ b/oidc-controller/api/core/oidc/issue_token_service.py
@@ -1,15 +1,15 @@
+import canonicaljson
 import dataclasses
 import json
 from datetime import datetime
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Optional, TypedDict
 
 import structlog
 from oic.oic.message import OpenIDSchema
 from pydantic import BaseModel
 
 from ...authSessions.models import AuthSession
-from ...verificationConfigs.models import ReqAttr, VerificationConfig
-from ..models import RevealedAttribute
+from ...verificationConfigs.models import AttributeFilter, VerificationConfig
 
 logger = structlog.getLogger(__name__)
 
@@ -20,6 +20,18 @@ PROOF_CLAIMS_ATTRIBUTE_NAME = "vc_presented_attributes"
 class Claim(BaseModel):
     type: str
     value: str
+
+
+# Used as a work around since these are represented as dictionaries in this case
+class ReqAttrDict(TypedDict, total=False):
+    names: List[str]
+    label: Optional[str]
+    restrictions: List[AttributeFilter]
+
+
+class RevealedAttributeDict(TypedDict, total=False):
+    sub_proof_index: int
+    values: dict
 
 
 class Token(BaseModel):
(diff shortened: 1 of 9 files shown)


oidc-controller/api/core/oidc/issue_token_service.py
```diff
@@ -22,6 +22,18 @@ class Claim(BaseModel):
...
+# Used as a work around since these are represented as dictionaries in this case
+class ReqAttrDict(TypedDict, total=False):
+    names: List[str]
+    label: Optional[str]
+    restrictions: List[AttributeFilter]
+
+
+class RevealedAttributeDict(TypedDict, total=False):
+    sub_proof_index: int
+    values: dict
```
esune, line 34: Wondering if these belong to a `models` source file, rather than in here. Thoughts? Additionally, since this is extremely similar to the model in `api/verificationConfigs/models.py` would there be a way of making things work without two type definitions (that then need to be maintained)?

esune (commented):
Looking good, just a question regarding the new models

oidc-controller/api/core/oidc/issue_token_service.py
Gavinok: I was able to adapt the code to use the existing version of ReqAttr and migrated RevealedAttribute to simply be a TypedDict since it is only used here and if we decided to include it in a Pydantic class this is supported. Pydantic just doesn't support inheriting from both BaseModel and TypedDict since BaseModel it's self does not inherit from TypedDict.

esune (approved):
👍🏻

PR #406, opened 2024-01-16
Handle Declined proof (abandoned state callback)

loneil:
**NOTE:** For this to function a subsequent upcoming change to ACA-Py will need to be released. ACA-Py currently does not send a problem report for abandoned presentations on connectionless presentation requests. The change to remove this restriction will need to be accepted and released.

However, on the VCAuthn side, this is just adding a new state handler, so this could conceivably be merged in without any adverse affects, the state just won't be triggered unless the ACA-Py change is in place.

---

Add a state handler in the controller acapy handler for a `present_proof` topic that contains a state `abandoned` (which is triggered when the user declines a presentation request).

The topic handler in the existing code already checks the incoming webhook `presentation_exchange_id` to get the appropriate auth session. So add a new check for abandoned state and emit that to the handler to display the HTML page with the new state. 
The state is just "abandoned", the error message says `abandoned: Declined` so possibly could parse out more if there were multiple `abandoned` cases to handle... but that would be relying on a text error message display rather than state.

See object contents from debug session below:
image

Alter the frontend page to handle this state. It looks like below and the user can hit try again or refresh to do a proof again if they accidentally decline. Can adjust wording/icon/color if needed.

image

Additionally refactor the JS in the HTML file to generalize the UI element setup and reduce duplication.

### DIFF
--- a/oidc-controller/api/templates/verified_credentials.html
+++ b/oidc-controller/api/templates/verified_credentials.html
@@ -136,6 +136,7 @@
       .header-desc.success,
       .header-desc.pending,
       .header-desc.expired,
+      .header-desc.abandoned,
       .header-desc.failed {
         display: none;
         border-radius: 0.5rem;
@@ -165,6 +166,9 @@
       }
       .header-desc.success {
         background-color: #dff0d8;
+      }
+      .header-desc.abandoned {
+        background-color: #f2dede;
       }
       .header-desc a {
         line-height: 1rem;
@@ -215,6 +219,18 @@
             >
           </div>
         </div>
+
+        <div class="header-desc abandoned">
+           <div class="qr-code-image">{{add_asset("circle-x.svg")}}</div>
+           <div class="qr-code-desc">
+             <b>Proof declined</b>
+             </br>
+             <a href="javascript:window.location.reload(true)"
+              title="Refresh QR code."
+               >Try again.</a
+             >
+           </div>
+         </div>
 
         <div class="header-desc pending">
           <div class="qr-code-image">{{add_asset("hourglass.svg")}}</div>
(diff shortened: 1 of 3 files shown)


oidc-controller/api/templates/verified_credentials.html
```diff
@@ -273,93 +289,27 @@ <h1>Scan with a Digital Wallet</h1>
     socket.connect();
 
     const toggleState = (state) => {
-      const intro = document.querySelector(".intro");
```
loneil, line 276: I refactored all this away to the setUiElements function below to generalize.

oidc-controller/api/templates/verified_credentials.html
```diff
@@ -273,93 +289,27 @@ <h1>Scan with a Digital Wallet</h1>
...
-
-          // Button elements
-          refreshButton.style.display = "flex";
-          scannedMask.style.display = "none";
-
-          // Turn off the spinner
-          qrcode.classList.remove("pending");
+          setUiElements(".expired", true, false, false);
+          break;
+        case "abandoned":
```
loneil, line 311: New state

loneil:
Draft ACA-Py changes can be seen in my fork here at this time https://github.com/loneil/aries-cloudagent-python/pull/1

Jsyro (approved):
Looks Great!

esune (approved):
LGTM 👍🏻

esune:
@loneil just confirming we can merge and the current code will not break - nothing will actually change until the related fix in ACA-Py is pulled in by a new image.

loneil:
@esune Yes it can go in, the only change depends on a state that won't show up yet so it won't have any affect on anything yet (I did refactor a bit on the UI page but tested out the other cases and there weren't any logic changes)

loneil:
Will try this out with the 0.12.0 RC0 acapy image soon on this PR.

loneil:
This was merged so put the 12.0 (tested and working locally) here https://github.com/bcgov/vc-authn-oidc/pull/407

esune:
> This was merged so put the 12.0 (tested and working locally) here #407

Sorry, I got merge happy today 🙂

loneil:
Probably best to keep it separate since it's an RC anyways

sgroh:
Sorry I got lost in messages, what is the ACA-Py version that could interoperate with this PR? I mean Once ACA-Py add the new "state",  Are we be able to use an external wallet like  eSatus/Lissi ?

PR #410, opened 2024-01-29
Updated ngrok configuration

### LINKED ISSUE #402: ngrok now requires a free account and authtoken
PM from ngrok here. I'm opening this issue to let you know what ngrok has recently started requiring a free account and authtoken to start the agent. It looks like your software already works with an authtoken, but there might be users or integrations or docs that need to updated.

Please reach out to support@ngrok.com if you are having any issues. 

Thanks again for using ngrok!

I've copied the original notice posted to twitter below:
```
As a part of our ongoing efforts to make ngrok more secure, starting on December 19, 2023, all usage of ngrok will require an account on either our free or paid tiers. 

To connect to the ngrok service via the agent, SDK, or ingress controller, you must add a valid authtoken that is linked to an active free or paid account. Any new or existing connections without a valid authtoken will result in error code 4018.

Learn more here: https://ngrok.com/docs/errors/err_ngrok_4018/
```

esune:
Resolves #402 

This PR updates the ngrok configuration to use a single container and therefore requiring only one authtoken to be provided.

To do so, a lot of the complexity in the `manage` script was removed, in particular options to run in multi-tenant mode have been removed since they made the script very complicated to maintain. It is still possible to run the service using a Traction agent, but this is not supported in the demo now.

Lastly, a tweak to the stale PR criteria was added to NOT mark as stale PRs that deal with dependency updates.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -18,4 +18,4 @@ jobs:
           stale-issue-label: 'stale'
           exempt-issue-labels: 'pinned,security,dependencies,epic'
           stale-pr-label: 'stale'
-          exempt-pr-labels: 'awaiting-approval,work-in-progress'
+          exempt-pr-labels: 'awaiting-approval,work-in-progress,dependencies'
--- a/docker/docker-compose-ngrok.yaml
+++ b/docker/docker-compose-ngrok.yaml
@@ -1,23 +1,18 @@
 version: "3"
 services:
-  controller-ngrok:
+  ngrok:
     image: ngrok/ngrok
     environment:
-      - CONTROLLER_SERVICE_PORT=${CONTROLLER_SERVICE_PORT}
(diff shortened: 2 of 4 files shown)


Jsyro (approved):
Simplifying the demo seems like a good idea. as long as there is some way to tests the controller config that depends on if it's a single-tenant or multi-tenant acapy.

esune:
> Simplifying the demo seems like a good idea. as long as there is some way to tests the controller config that depends on if it's a single-tenant or multi-tenant acapy.

Unit tests or manual configuration are still supported, however it was just too complicated to maintain and most use-cases will likely run a read-only agent. That being said, vc-authn can still be ocnfigured to use a Traction tenant, and I believe unit-tests cover that bit too (will check and log issues if necessary).

PR #411, opened 2024-02-04
Update demo app libraries

loneil:
The Demo app has a number of dependabot warnings. Update to clear those and get more up to date on lib versions.

Doesn't update to Vue 3, that would be a bigger exercise (should be simple enough, there's an equivalent Vue 3 Scaffold we'd made that is the same setup as this Vue 2 app) but should be done at some point as Vue 2 is end of life.

Built and ran demo app with VCAuthN locally and can log in and use the app as expected with a cred after this update.

Though the call to the backend to make a secured call doesn't seem to work. Not sure how long that's been an issue (or if ever worked) but it's the same before the updates from this PR. Could look into at some point

image

### DIFF
--- a/demo/vue/app/frontend/package-lock.json
+++ b/demo/vue/app/frontend/package-lock.json
@@ -9,16 +9,16 @@
       "version": "1.0.0",
       "license": "Apache-2.0",
       "dependencies": {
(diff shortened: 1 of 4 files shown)


esune (approved):
Looks good! I believe that call never worked, would be good to get it sorted out.

Issue #413, opened 2024-02-08
Deep Link button on QR code page


esune:
We need a deep-link button below the QR Code to allow same-device authntication.
The deep link should be i nthe format `bcwallet://aries_proof-request?_oob=...`

loneil:
@esune wallet OOB handling issue we discussed laid out here

https://github.com/bcgov/bc-wallet-mobile/issues/953#issuecomment-1947038869

knguyenBC:
I did some quick wireframes to help improve the UI of the screen based off of recent user research. https://xd.adobe.com/view/b40869d2-5fff-40ff-855c-ace0d1da0c30-e15b/ explanation of design choices in the comments.

loneil:
Excellent! Thanks so much @knguyenBC

loneil:
From discussion today with @esune 
Plan to handle generating the correct payload for the deep link and accommodate for both link and QR. Normally this payload generation is not done until the short QR link is redirected to.

1. Access the page first load
1. It generates a connectionless or oob (config flag) payload that looks like this (depending on type)
![image](https://github.com/bcgov/vc-authn-oidc/assets/17445138/0400466c-ca84-488b-9511-82438e6bb82d)

1. It stores that in the database
1. It generates the short link and QR code
1. It returns that to the screen in the Deep Link Button
1. It also returns the QR code

**On deep link click**
- just action that link

**On QR scan**
1. Follow redirect short link
1. In handler, LOAD that from the DB instead of generating it
1.  Return same as before

**On refresh timer**
1. Generate that again and overwrite the stored version
1. return deep link and qr again to the page

PR #419, opened 2024-02-15
Display pres req link for debugging

loneil:
(this is not related to BC Wallet deep link feature upcoming)

It's onerous to debug the QR code link on a VCAuth login. Have to snip code out, save, upload to a qr decode or something like that. Would be good to just see the pure text link that's converted to the QR Code.

Can use this to debug locally, or if needed do operational support for a user that's having an issue with the QR code.

Add an easter egg (BC Wallet style, tap 10 times on the BC Gov logo) that exposes the link in a textbox

image

### DIFF
--- a/demo/vue/app/frontend/src/router/index.js
+++ b/demo/vue/app/frontend/src/router/index.js
@@ -88,7 +88,7 @@ export default function getRouter(basePath = '/') {
         document.title = to.meta.title
           ? to.meta.title
           : process.env.VUE_APP_TITLE;
-      } else document.title = 'Demo VC-Authn-OIDC App”'; // default title
+      } else document.title = 'Demo VC-Authn-OIDC App'; // default title
 
       if (to.query.r && isFirstTransition) {
         router.replace({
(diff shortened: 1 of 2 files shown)


swcurran:
There is another issue (#413 ) that requires we add a deep link button to the VC-Authn-OIDC.  That ticket should be a very high priority.  This implements a lot of it and happy to merge it, but please take a look and carry out that issue.

loneil:
https://github.com/bcgov/vc-authn-oidc/issues/413 is in progress

demo/vue/app/frontend/src/router/index.js
```diff
@@ -88,7 +88,7 @@ export default function getRouter(basePath = '/') {
         document.title = to.meta.title
           ? to.meta.title
           : process.env.VUE_APP_TITLE;
-      } else document.title = 'Demo VC-Authn-OIDC App”'; // default title
```
jamshale, line 91: Think I did this a long time ago :/

jamshale (approved):

esune (approved):

PR #421, opened 2024-02-16
BC Wallet deep link display

loneil:
Add capability of displaying a deep link specific to the BC Wallet app. On devices that can have the app (iOS, Android) show the Deep link as the main option, with an ability to get a QR code as well.
See Kim's mockups here for UI/UX https://xd.adobe.com/view/b40869d2-5fff-40ff-855c-ace0d1da0c30-e15b/

Additionally, change the flow to generate the presentation request message contents when the controller loads (and creates the QR code and deep link), rather than waiting until the short QR link is called back to.
This message content is stored with the auth_session document in the mongo collection, so that it is _fetched_ instead of generated on that QR callback.

### DIFF
--- a/oidc-controller/api/authSessions/models.py
+++ b/oidc-controller/api/authSessions/models.py
@@ -1,6 +1,6 @@
 from datetime import datetime, timedelta
 from enum import StrEnum, auto
-from typing import Dict
+from typing import Dict, Optional
 
 from api.core.acapy.client import AcapyClient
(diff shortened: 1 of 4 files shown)


esune (dismissed):
Looks good based on what was reviewed with @loneil .
Once merged we can create a pre-release for `dev` and test in a live environment.

esune (approved):

Issue #425, opened 2024-02-22
Need index on presentation exchange ID


loneil:
Documents from the oidccontroller.auth_session collection are fetched by `pres_exch_id` and there are no indexes for that presently. This could affect performance of VCAuthn, especially with the considerations here https://github.com/bcgov/vc-authn-oidc/issues/424

There is also a get_by_pyop_auth_code that is called to fetch by the `pyop_auth_code` field that would likely need indexing as well

Example

![image](https://github.com/bcgov/vc-authn-oidc/assets/17445138/25c12491-2659-40fb-8595-721842c3dd55)

![image](https://github.com/bcgov/vc-authn-oidc/assets/17445138/9e12b5aa-f1a2-4d33-b897-88aa68d03c08)

loneil:
Proposed change and discussion here
https://github.com/bcgov/vc-authn-oidc/pull/426

loneil:
Completed a while back

PR #426, opened 2024-02-23
Add indexes

loneil:
https://github.com/bcgov/vc-authn-oidc/issues/425

Add indexes for the fields `pres_exch_id` and `pyop_auth_code` for the `auth_session` collection.

Those fields are used to look up on at various points in the VCAuthn flow so they should need indexing.
Can look for `get_by_pyop_auth_code` and `get_by_pres_exch_id` for usage in the code.

This just uses the pymongo `create_index` command in the init_db part of the app, which is called on startup of the oidc-controller.
This successfully added indexes on my local where there are around 500 documents already populated. 
**Note for consideration**: do we need to think about index addition for running VCAuthn instances that may(?) have a **lot** of documents in the collection? Would that need millions? Billions? in order to be concern with an index addition? Not sure if we are close to that in any existing VCAuthn contexts using the `auth_session` collection.

Result right after adding and restarting controller:

image


Jsyro (approved):
Looks good!

jamshale:
Definitely seems like the correct indexes.

jamshale (approved):

esune:
> Definitely seems like the correct indexes.
> 
> What is the `auth_code`? This is a unique identifier for the login? Or is this common number code? It won't help and might hurt if a lot of the entries are the same.
 
The `auth_code` is uniquer per authentication attempt, it won't be shared by sessions so it is safe to have an index on it if we are querying that collection based on its value.

esune (approved):

PR #436, opened 2024-03-15
Use local did for read-only agent

esune:
Add the `--wallet-local-did` startup parameter to prevent the read-only agent from attempting to write their DID to the ledger.
I am unsure on how this issue was triggered, however looking at the ACA-Py documentation I think this is the correct fix - see https://github.com/hyperledger/aries-cloudagent-python/issues/2828

The chart version has been updated in preparation for the next release as well.


loneil (approved):
Ran locally with this change in docker and it works now.
Helm changes look correct 👍

Issue #432, opened 2024-03-05
Switch to using embedded Vue app for QR Code page


esune:
The QR code page is fairly simple, and was therefore developed with plain JS. This is however causing challenges with the handling of websockets, among other things, as they are being interrupted when the phone switches to BC wallet from the browser (deep-linking scenario).

using an embedded Vue app will allow us to improve handling of these events without relying on custom code, as well as allow for future enhancements/improvements.

c.c.: @loneil @Gavinok

esune:
Example: old VC-AuthN authorize page https://github.com/bcgov/vc-authn-oidc/blob/v1.0.0/oidc-controller/src/VCAuthn/Views/Authorize/Authorize.cshtml

esune:
One thing to check-out while restructuring the app using Vue: when BC Wallet scans the QR Code, in some scenarios (i.e.: using OOB) the QR code scans, but BC Wallet says it is invalid and doesn't move forward. At that point, it's a dead-end in VC-AuthN since there is no way to refresh the page (other than refreshing through the browser) and the proof-request will not timeout as the QR code was scanned. We may want to add a control to start over the process in these scenarios, like the refresh control that appears when the proof has expired.

loneil:
Yeah good call, some sort of "Taking too long? Try again" status display on a (longer) timer or something.

esune:
> Yeah good call, some sort of "Taking too long? Try again" status display on a (longer) timer or something.

Yes, that would take care of that scenario AND potentially others with similar interaction patterns

Gavinok:
Sounds good I'll add that now. Thanks for the ideas!

PR #441, opened 2024-03-19
Migrate to bootstrap

### LINKED ISSUE #431: Transition to Bootstrap for UI positioning
The current UI has a significant amount of CSS for handlign the layout of elements. Transitioning to using Bootstrap Grid will simplify the CSS as well as make the project more maintainable.

c.c.: @loneil @Gavinok

Gavinok:
This pr resolves #431

All that remains when it comes to CSS are classes or IDs accessed via javascript (should be resolved with #432) or custom overrides for bootstrap it's self. 

Updated look for desktop:

Updated look for mobile:

### DIFF
--- a/oidc-controller/api/templates/verified_credentials.html
+++ b/oidc-controller/api/templates/verified_credentials.html
@@ -139,19 +139,19 @@
         height: 2rem;
         opacity: 0.7;
       }
-      .expired {
+      .header-desc.expired {
         background-color: #f0f0f0;
       }
-      .failed {
+      .header-desc.failed {
         background-color: #f2dede;
       }
-      .pending {
+      .header-desc.pending {
         background-color: #cbe8ff;
       }
-      .success {
+      .header-desc.success {
         background-color: #dff0d8;
       }
-      .abandoned {
+      .header-desc.abandoned {
         background-color: #f2dede;
       }
       .disabled {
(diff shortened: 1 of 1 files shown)


Gavinok:
I will setup a meeting to discuss any changes to be made

loneil:
One note from Kim, can we move the refresh icon to the left here
image

like it was originally
image
(but keeping the new opaqueness, that's a good improvement for readability)

"To the left please. That makes it seem more like a button and less like a status"

loneil:
Also get a big blue outline in the pending state

image

loneil:
Just a couple tweaks I noticed above that are worth focusing on but in general looking good.

I think there's some more grid changes and alignment tweaks for the "header-desc" info boxes we should do, however not worth looking at right now as they are working close enough as intended and we should go through a UX review to see if those will dramatically change anyways.

So (after tweaks above) worth merging and deploying to dev (to check on other devices) and then having UX review with Kim before going too much further

Gavinok:
> Just a couple tweaks I noticed above that are worth focusing on but in general looking good.
> 
> I think there's some more grid changes and alignment tweaks for the "header-desc" info boxes we should do, however not worth looking at right now as they are working close enough as intended and we should go through a UX review to see if those will dramatically change anyways.
> 
> So (after tweaks above) worth merging and deploying to dev (to check on other devices) and then having UX review with Kim before going too much further

Just pushed these changes. Thanks for the feedback and testing. Should be good to go now.

loneil (approved):
Looks good for this milestone.

PR #448, opened 2024-03-27
Updated version of FastAPI (migrate to Pydantic2 and others)

loneil:
To cover FastAPI security updates (new versions of FastAPI use Pydantic 2) and future-proofing, need to update the oidc controller in VCAuthn to use Pydantic 2. Other minor FastAPI update tweaks required as well.

Some relevant Pydantic2 changes

Optional logic changed in Pydantic 2
https://docs.pydantic.dev/latest/migration/#required-optional-and-nullable-fields

Using ConfigDict instead of class based config
https://docs.pydantic.dev/latest/migration/#changes-to-config

Get validators is gone
https://docs.pydantic.dev/latest/migration/#defining-custom-types

Additionally needed minor tweak to the way the socket.io endpoint sets up as the mount to `/ws` stopped working in newer FastAPI (or python-socket) so defining the whole path in the socket app works now (explicitly set `socketio_path`).


esune (approved):
LGTM 👍🏻

PR #449, opened 2024-03-28
Move to using embedded Vue app for QR Code page

### LINKED ISSUE #432: Switch to using embedded Vue app for QR Code page

Gavinok:
This pr resolves #432

In addition to the migration to Vue this PR adds a timer to catch when a credential has been stuck in the pending state for over a minute prompting the user.

As well as providing this feedback users are also informed on possible reasons for a failed credential.

I am not 100% sure these error messages make sense in all cases so feedback on them is appreciated.


oidc-controller/api/templates/verified_credentials.html
```diff
@@ -333,193 +287,237 @@ <h1 class="mobile-device my-1">Continue with:</h1>
...
-        deepLink.href = "";
-      }
+     createApp({
+       data() {
+         const mobileDevice = ref(
+           getBrowser() === "Android" || getBrowser() === "iOS" ? true : false
+         );
+      const errorMessage = computed(() => {
+      if (state.current === "failed") {
+      return `<b>We couldn't verify your credentials because they may be:</b>
```
amanji, line 340: Can these not all be Vue components themselves, now?

Gavinok: Couldn't agree more. Once I'm back I will extract it. For whatever reason I couldn't find a way to do it without multiple files or a build step but I was able to dig up a solution.

oidc-controller/api/templates/verified_credentials.html
```diff
@@ -333,193 +274,261 @@ <h1 class="mobile-device my-1">Continue with:</h1>
...
-    });
+    </div>
 
-    socket.on("connect", () => {
-      socket.emit("initialize", { pid: "{{pid}}" });
-    });
+    <script src="https://unpkg.com/vue@3/dist/vue.global.js"></script>
+  </body>
+  <!-- Components -->
+  <script type="text/x-template" id="status-description">
```
amanji, line 287: Can these not be imported from separate files? Just taking a brief look at: https://vuejs.org/guide/extras/web-components.html#sfc-as-custom-element if there's any relevance here.

Gavinok: This was more a requirement we went with to simplify the situation when it came to how the template is used by the python code. We definitely could split it off but would need to handle the template expansion aka all the `{{add_asset("hourglass.svg")}}` which are being expanded by https://github.com/bcgov/vc-authn-oidc/blob/embedded-vue/oidc-controller/api/templates/helpers.py#L2

Gavinok: Correct me if I am wrong @loneil @esune . I easily could have misunderstood the plans for this.

Gavinok: I could try and change [get_authorize](https://github.com/bcgov/vc-authn-oidc/blob/embedded-vue/oidc-controller/api/routers/oidc.py#L86-L212) to support multiple files.

loneil: The goal is to have this be contained and served in the html file so that shouldn't entail having to have anything built with Vite/etc or using `.vue` SFCs that need building.

amanji: Ok In that case this seems like a good approach.

amanji (dismissed):

loneil (changes requested):
Not seeing the redirecting working when trying locally any more.

It's sitting on 
image

and I don't see `callback_url` referenced in the code any more

Gavinok:
> Not seeing the redirecting working when trying locally any more.
> 
> It's sitting on image
> 
> and I don't see `callback_url` referenced in the code any more

Just tested this. Should work now

loneil (approved):
Looks good after redirect fix. Tested out browser and mobile flows.

esune (approved):

Issue #429, opened 2024-03-04
CD pipeline for dev


esune:
Implement a CD pipeline to deploy the latest changes to `dev` every time something is merged to the `main` branch.

esune:
For reference, the `traction` CD pushing to `dev` every time a change is merged to `main` is here: https://github.com/bcgov/traction/blob/main/.github/workflows/on_push_main.yaml

We would likely want something similar implemented.

esune:
There's a couple last tweaks required: pipeline is now working appropriately, however the deployment is trying to pull the `dev` image, which does not exist. @i5okie pointed out it would be best to just use the sha of the built image.

We need to replace the image tag in the `helm upgrade` command to use that rather than `dev`.

PR #487, opened 2024-04-22
Add workflow for auto deployment to dev for main branch changes

Gavinok:
This is a draft for #429

This will likely need extra review since there isn't an option for local testing in this case. Most of the code has been ported from 

https://github.com/bcgov/traction/blob/restore-json-toggle-schema/.github/workflows/on_push_main.yaml


i5okie (commented):
Have the secrets been added in the repo settings?

i5okie (dismissed):
lgtm

Gavinok:
> Have the secrets been added in the repo settings?

As far as I can tell I don't have the proper permissions to add or see them. Since we have not workflows using them yet they probably need to be added

esune:
@i5okie we added all certificates except for `OPENSHIFT_CA_CRT` - I am not 100% sure what should go there. Can you help with that?
After adding that secret we should be able to merge and test this.

esune (approved):
👍🏻

.github/workflows/on_push_main.yaml
```diff
@@ -0,0 +1,47 @@
...
+          namespace: ${{ secrets.OPENSHIFT_NAMESPACE }}
+
+      - name: VC-Authn Dev Helm
+        run: |
+          helm upgrade --install vc-authn-oidc \
+          -f ./charts/vc-authn-oidc/templates/deployment.yaml \
+          --set image.tag="dev" \
+          ./charts/vc-authn-oidc --wait
+      - name: Restart Deployments
+        # TODO determine the vc-authn equivalent
```
loneil, line 44: Should the TODO comments stay in?

esune: Ha, good catch! It looks like this is resolved so we should remove it

.github/workflows/on_push_main.yaml
```diff
@@ -0,0 +1,45 @@
...
+      - name: VC-Authn Dev Helm
+        run: |
+          helm upgrade --install vc-authn-oidc \
+          -f ./charts/vc-authn-oidc/templates/deployment.yaml \
+          --set image.tag="dev" \
+          ./charts/vc-authn-oidc --wait
+      - name: Restart Deployments
+        run: |
+          oc rollout restart deployment/vc-authn-oidc
+          oc rollout restart deployment/vc-authn-controller
```
esune, line 45: I double checked, we actually need to trigger deployments `vc-authn-oidc-agent` and `vc-authn-oidc`

esune (changes requested):

.github/workflows/on_push_main.yaml
Gavinok: Oh, I'll add that now. My mistake, I figured the agent was fine.

.github/workflows/on_push_main.yaml
```diff
@@ -0,0 +1,46 @@
...
+        run: |
+          helm upgrade --install vc-authn-oidc \
+          -f ./charts/vc-authn-oidc/templates/deployment.yaml \
+          --set image.tag="dev" \
+          ./charts/vc-authn-oidc --wait
+      - name: Restart Deployments
+        run: |
+          oc rollout restart deployment/vc-authn-oidc-agent
+          oc rollout restart deployment/vc-authn-oidc
+          oc rollout restart deployment/vc-authn-controller
```
esune, line 46: This actually doesn't exist, we should remove it completely

esune (changes requested):

PR #489, opened 2024-04-25
Fix CD pipeline definition

esune:
Fix issues with CD pipeline


Gavinok (approved):
Thanks for the corrections

PR #492, opened 2024-04-25
Added outputs to publish action for use in on_push_main

Gavinok:
This should resolve the remaining issue mentioned in #429

While this does expose the image version something I am concerned about is that the tag input is possibly not being used at all in https://github.com/bcgov/vc-authn-oidc/blob/use-dev-image-in-helm-command/.github/workflows/publish.yml

We  can see this in https://github.com/bcgov/vc-authn-oidc/actions/runs/8825653481/job/24230180441 Build VC-AuthN / Publish VC-AuthN Image as opposed to Build VC-AuthN / Publish VC-AuthN dev Image as you would expect based on this expression
https://github.com/bcgov/vc-authn-oidc/blob/use-dev-image-in-helm-command/.github/workflows/publish.yml#L2

### DIFF
--- a/.github/workflows/on_push_main.yaml
+++ b/.github/workflows/on_push_main.yaml
@@ -37,7 +37,7 @@ jobs:
         run: |
           helm upgrade --install vc-authn-oidc \
           -f https://raw.githubusercontent.com/bcgov/trust-over-ip-configurations/main/helm-values/vc-authn-oidc/dev.yaml \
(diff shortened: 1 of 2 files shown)


esune (approved):
LGTM, waiting for a second pass from @i5okie before merging

i5okie (approved):
Looks good to me.

PR #500, opened 2024-04-30
Library updates

loneil:
Cover the dependabot warnings and additional library updates (see changes) in 
- OIDC Controller
- OIDC Controller dev dependencies
- Vue demo


loneil:
Can hold on merge as needed if any additional work on the workflow is needed but it looks like that's working to me.

Gavinok (approved):
👍

PR #519, opened 2024-05-14
Set Pending state on Deep Link action

loneil:
The Deep Link when clicked was opening the pres req on the BC Wallet but leaving the button enabled. This can cause the issue discussed here
https://github.com/bcgov/bc-wallet-mobile/issues/1956#issuecomment-2105385631

Change it to set the deep link button to disabled (Pending status triggered manually, since it's not updating on the server when deep link is used, no callback)

See comments for relevant changes in html file, diff is big because there was a extraneous div ending that was breaking prettier so I fixed that and formatted.


oidc-controller/api/templates/verified_credentials.html
```diff
@@ -181,24 +183,26 @@
...
           <div v-if="mobileDevice && !failed">
             <a
               id="deep-link-button"
               href="{{wallet_deep_link}}"
-              type="submit"
               class="btn btn-lg btn-block btn-primary w-100 my-2 rounded-1"
               title="Open BC Wallet"
               target="_blank"
               :class="{disabled: state.disableDeeplink}"
+              v-on:click="deepLinkOpen"
```
loneil, line 205: Add a click handler that sets up the state.
Leave the href and target blank rather than doing a window nav in the `deepLinkOpen` function. Doing it this way prevents an "allow this to open" confirmation that happens if it's done in JS.

oidc-controller/api/templates/verified_credentials.html
```diff
@@ -326,211 +331,224 @@ <h1 class="my-2">
...
+          displayQr: mobileDevice.value ? false : true,
+          svgContent: `{{add_asset("BCID_H_rgb_rev.svg")}}`,
+          bcIdClicks: 0,
+          socket: null,
+        };
+      },
+      methods: {
+        deepLinkOpen() {
+          state.disableDeeplink = true;
+          this.setUiStates("pending");
```
loneil, line 391: Set Deep Link pending status

oidc-controller/api/templates/verified_credentials.html
```diff
@@ -326,211 +331,224 @@ <h1 class="my-2">
...
+          state.showScanned = ["pending", "verified"].includes(status);
+          state.showRefresh = ["abandoned", "expired", "failed"].includes(
+            status
+          );
+          state.disableDeeplink = [
+            "abandoned",
+            "expired",
+            "pending",
+            "success",
+          ].includes(status);
```
loneil, line 412: Refactor to be called from multiple spots

oidc-controller/api/templates/verified_credentials.html
```diff
@@ -326,211 +331,224 @@ <h1 class="my-2">
...
-         state.current = data.status
-         state.showSpinner = data.status === "pending";
-         state.showScanned = ["verified", "pending"].includes(data.status);
-         state.showRefresh = ["failed", "expired", "abandoned"].includes(data.status);
-         state.disableDeeplink =  ["expired", "success"].includes(data.status)
-         console.log("Toggling state", this.state);
-       });
+        this.socket.on("status", (data) => {
+          console.log("Toggling state", data.status);
+          this.setUiStates(data.status);
```
loneil, line 430: Refactor to be called from multiple spots

esune (approved):

PR #522, opened 2024-05-17
[Helm} Chart changes to support ArgoCD deployment

### LINKED ISSUE #447: Update Helm Chart to add ArgoCD deployment compatibility
The helm chart must be updated to be deployed with ArgoCD.
ArgoCD uses `helm template` command when deploying, rendering logic, like secret lookup inoperable. If deployed as is with ArgoCD, any existing secret will be overwritten.

i5okie:
- Add support for existing secrets
- Address database storage limitations (https://github.com/bcgov/trust-over-ip-configurations/issues/200)
- Upgrade MongoDB version

Resolves #447


esune (dismissed):
LGTM 👍🏻 

Do we need to update anything [here](https://github.com/bcgov/trust-over-ip-configurations/blob/main/helm-values/vc-authn-oidc/dev.yaml) as well?

i5okie:
> LGTM 👍🏻
> 
> Do we need to update anything [here](https://github.com/bcgov/trust-over-ip-configurations/blob/main/helm-values/vc-authn-oidc/dev.yaml) as well?

Yes, values will need to be updated as well.

esune (approved):
Changes looking good to me. I assume they have been tested throughly already?

loneil:
Think I'm reading the Bitnami chart correctly, this is going to Mongo 7.0.9 right? Just that local docker compose is on `moongo:6.0` so we could update and test that as well

PR #531, opened 2024-05-22
auth_session document clean up

### LINKED ISSUE #424: Consider mongo DB auth_session document clean up
For each presentation request that is created an auth_session document is created in the Mongo store.
As the page is left open, the refresh of expired requests creates a new one each refresh as well.

If a VC Authn instance gets very high usage (or people leave the page open forever) this collection could get large. While Mongo should be able to handle this, we might have concerns about PVC size blowing up keeping all these records around.

Could consider
1. This is not a problem, just let disk space get high (they are small records so it would need lots of usage)
2. Instead of keeping the `expired` proof_status ones around delete them? (might not be accommodated on the frontend...) But then there's still all the `verified` and other status once staying around forever
3. Some cleanup job?

Gavinok:
This pr resolves #424 

This makes use of MongoDB's TTL index which will delete expired entries https://www.mongodb.com/docs/manual/tutorial/expire-data/ . 

In this case we set the TTL based on CONTROLLER_PRESENTATION_EXPIRE_TIME as well as a new CONTROLLER_PRESENTATION_BUFFER_TIME which is used to ensure that we do not delete an auth session too soon before the UI can update. 

As of now we have a TTL index that will delete expired, failed, and abandoned auth_sessions.

Note that the expireAfterSeconds does guarantee that it will be deleted after that many seconds. Instead that it will not be deleted before the given number of seconds have passed.


oidc-controller/api/db/session.py
```diff
@@ -23,6 +25,17 @@ async def init_db():
     auth_session.create_index([("pres_exch_id", ASCENDING)], unique=True)
     auth_session.create_index([("pyop_auth_code", ASCENDING)], unique=True)
 
+    expire_time: int = settings.CONTROLLER_PRESENTATION_EXPIRE_TIME + settings.CONTROLLER_PRESENTATION_BUFFER_TIME
+
+    for k, v in [("expired_ttl", AuthSessionState.EXPIRED),
+                 ("failed_ttl", AuthSessionState.FAILED),
+                 ("abandoned_ttl", AuthSessionState.ABANDONED)]:
```
esune, line 32: Is there a way we could provide this in a configuration file so that behaviour could be customized per-deployment? In that case, how would we ensure that indexes that have been removed from the config are removed from the mongodb TTL settings as well?

esune (commented):
I like the approach, this reduces the amount of code we need to maintain to handle this by quite a bit 👍🏻 

The question I have is on whether this can be configured rather than hard-coded.

oidc-controller/api/db/session.py
Gavinok: Likely yes. Let me look through the documentation on mongoDB's JS interface

Gavinok: I guess you probably mean a dedicated config file. That is also possible (and probably preferable)

Gavinok: I have moved this to a json config file

oidc-controller/api/authSessions/sessiontimeout.json
```diff
@@ -0,0 +1,5 @@
+{
+	"expired": 10,
+	"failed": 10,
+	"abandoned": 10
+}
```
loneil, line 5: Do these get overridden by environment variables or helm values?
Or is this hard-coding, just in a json file that gets imported? In that case I'd just make them constants in the code to remove the extra file and the whole file import logic complication.

loneil (commented):
Not seeing the benefit of that JSON file and all the code to use it if it's not configurable?

loneil:
Will we leave `verified` state permanently? Or should that be flagged for deletion as well?

oidc-controller/api/authSessions/sessiontimeout.json
Gavinok: The idea was that users could modify the file to extend different expiration times and for example add expiration for verified auth sessions. It was meant to be used similar to https://github.com/bcgov/vc-authn-oidc/blob/autodelete-auth-session/oidc-controller/api/logconf.json

Gavinok:
> Will we leave `verified` state permanently (someone scans and shares successfully)? Or should that be flagged for deletion as well?

This is the point of the config file. I am open to suggestions on how we could make it more configurable than the file. I considered making at a REST API endpoint which was why I stuck with a json object.

oidc-controller/api/db/session.py
```diff
@@ -23,6 +28,32 @@ async def init_db():
...
+        try:
+            for k, v in experation_times.items():
+                assert isinstance(k, str)
+                assert k in auth_session_states
+                auth_session.create_index(
+                    [("created_at", ASCENDING)],
+                    expireAfterSeconds=v + settings.CONTROLLER_PRESENTATION_BUFFER_TIME,
+                    name=k + "_ttl",
+                    partialFilterExpression={"proof_status": {"$eq": k}},
+                )
```
loneil, line 47: Some thoughts on this (might change depending on the JSON file config pattern, TBD).
We're creating indexes in a loop and one per expiry condition, so we get multiple indexes:
image

Is the [partialFilterExpression](https://www.mongodb.com/docs/manual/core/index-partial/#create-a-partial-index) able to just take an OR condition instead? So there would be 1 index. Unless there's good reason to have multiple partial indexes.

If it **were** to just simplify to deleting all records regardless of state after a time it wouldn't have to be a partial index.

Also if using this pattern if one
- Applies the indexes to delete based on some config
- Changes the config
- Restarts the app
Will it then skip over changing the indexes? Because they are already there and it's only running `create_index` I don't _think_ it would remove previously created partials, or alter them?

Gavinok: Previously we couldn't use OR but now that we are on mongo 7 we actually can which is great news! I should have just made this dependant on the mongo update. I'll switch this to an or now. As of now it was a constant once the index is set it would need to be removed however I can add support for this while moving to the new use of `OR`. Thanks for mentioning this!

oidc-controller/api/db/session.py
```diff
@@ -23,6 +28,32 @@ async def init_db():
     auth_session.create_index([("pres_exch_id", ASCENDING)], unique=True)
     auth_session.create_index([("pyop_auth_code", ASCENDING)], unique=True)
 
+    with open(
+        (
+            Path(__file__).parent.parent / "authSessions" / "sessiontimeout.json"
```
esune, line 33: This path should be made an environment variable, so a file that is NOT in the application working folder can be passed. This can be the default value if nothing is provided.

oidc-controller/api/db/session.py
```diff
@@ -23,6 +28,32 @@ async def init_db():
...
+        ).resolve()
+    ) as user_file:
+        experation_times: dict[str, int] = json.loads(user_file.read())
```
esune, line 36: Nit.: typo, should be `expiration`

esune (changes requested):
I think the config file should work - especially with the new improved `OR` clause as suggested by @loneil . However, it should be a configuration overlayed at runtime rather than a file in the application context directory.

Please make it "mountable" like [this example](https://github.com/bcgov/vc-authn-oidc/blob/18af14e939843454db6e1211e4f6c073a7b73a07/docker/docker-compose.yaml#L124) and remember to update the Helm chart accordingly as well. We can provide reasonable default values to be used if nothing is specified.

Gavinok:
@loneil @esune I realized a limitation of moving to a single ttl index is that since we can only assign one expiration time per index we can't have a different expiration time for each proof_status. aka expired, aborted, failed, etc can only expire at one time unless we make multiple indexes.

 To me multiple indexes seems like the best option as long as we want to support multiple expiration times for each state.

Gavinok:
I have now updated this to support the following
- [X] updating indexes
  - [X] Create if it does not exist
  - [X] Update if it does
  - [X] If not in config delete it
- [X] Support restarting the service
- [X] Make the path to the config file an environment variable
- [-] Make the config file mountable
  - [X] https://github.com/bcgov/vc-authn-oidc/blob/18af14e939843454db6e1211e4f6c073a7b73a07/docker/docker-compose.yaml#L124
- [X] Add friendly defaults if the file can not be found
  - best default is to maybe not delete anything?
  
The only thing I was unsure of was how to reflect this mountable file in the helm charts.

Gavinok:
The following changes have taken place
- [X] single index for ttl
- [X] notify the user of the need to manually delete the index
- [X] no longer drop all indexes

oidc-controller/api/core/config.py
```diff
@@ -167,6 +167,15 @@ class GlobalConfig(BaseSettings):
...
 
+    # How long auth_sessions with matching the states in
+    # CONTROLLER_SESSION_TIMEOUT_CONFIG_FILE are stored for in seconds
+    CONTROLLER_PRESENTATION_CLEANUP_TIME: int = os.environ.get(
+        "CONTROLLER_PRESENTATION_CLEANUP_TIME", 86400
+    )
+
+    CONTROLLER_SESSION_TIMEOUT_CONFIG_FILE: Optional[str] = os.environ.get(
+        "CONTROLLER_SESSION_TIMEOUT_CONFIG_FILE"
+    )
```
esune, line 178: This assumes there MUST be a file with the configuration. Can we provide either a default file or a set of default values to use so the service can be run as-is?

esune (changes requested):
Only one comment about the default config file, changes look good otherwise.

We will need to update the documentation to indicate the new settings as well.

oidc-controller/api/core/config.py
Gavinok: A config file is not necessary the user will just be warned that no ttl will be applied. I can still add defaults though. I figured a safe default was not to delete anything by default but instead I can just make it use any non verified auth session.

oidc-controller/api/authSessions/sessiontimeout.json
esune: I should have read all the log statement rather than the first line only 🙂

oidc-controller/api/core/config.py
esune: I should have read past the first line of the log statement 🙂 
An update to the docs and we should be good then!

Gavinok: Just added documentation. Thanks

esune (dismissed):
LGTM 👍🏻

esune:
@Gavinok changes look good, but the tests are still failing - it seems like a linting error currently.

esune (approved):

PR #532, opened 2024-05-28
Update to mongo 7 on local Docker

loneil:
Use `mongo:7.0` tag in Docker Compose (tried locally with an existing setup).
Can merge in if we validate the bitnami update for Helm went all good.


Gavinok (approved):
Looks good

PR #546, opened 2024-06-13
feat: allow url override

### LINKED ISSUE #543: Ngrok blocked on BCNGN
Ngrok is blocked/band on BCNGN. May be handy to decouple from ngrok to so that `./manage start` works without ngrok. Perhaps just by defining some URLs like below. This way alternative methods to expose the services externally can be used.

```console
export VC_AUTHN_CONTROLLER_URL=
export VC_AUTHN_ACAPY_URL=
```

jleach:
Allow the external URLs to be overiden rather than relying on ngrok. This will allow devs who cannot use ngrok to user any means they like.

Fixes #543


esune (approved):
Thanks @jleach 👍🏻

Issue #504, opened 2024-05-04
Use URL instead of base64-encoded payload for deep link


esune:
When proof-requests grow in size, the resulting deep-link can grow to over 2048 characters, causing the payload to be truncated when sent over to BC Wallet (2048 characters is the size limit for URLs) which will then fail to process the request as the decoded output is a malformed JSON.

One way to fix this would be to use a URL/shortened URL instead of the base64 encoded presentation-request, but we are unsure on whether this will work out-of-the-box or not (attempts seem to indicate this will not work as-is).

@jleach @cvarjao does a change need to get into BC Wallet to handle this, or are we maybe using the `bcwallet://...?c_i=https://my-url.com` (or similar OOB) syntax incorrectly with the url?

esune:
Possibly superseeded by #513

cvarjao:
I think we need to talk more about invitation URLs and how payloads are retrieved. I still don't fully understand why we process HTTP Header (`Location`) redirect URL as opposed to just get the payload itself in the response body

esune:
> I think we need to talk more about invitation URLs and how payloads are retrieved. I still don't fully understand why we process HTTP Header (`Location`) redirect URL as opposed to just get the payload itself in the response body. And it needs to support deep link (custom protocols)

This issue was logged specifically for the issue relating to deep-link URLs exceeding the 2048 character limit. If we switch our pattern to using short-lived connections we should be able to also resolve this.

We may still want to chat as I am not 100% sure I follow the location redirect URL and custom protocols bit.

esune:
Thanks to the BC Wallet team for implementing the new deep-link handler - we should update our code to generate the deep link using the following format: `didcomm://?_url=<base64 encoded url>`.

loneil:
Handling here https://github.com/bcgov/vc-authn-oidc/pull/557
BC Wallet release with this is slated to be out next Monday, probably still worth having the config toggle in there though IMO for any needs to troubleshoot it back and forth to the older format or anything for a while.

PR #557, opened 2024-06-18
Support upcoming ?_url handler in wallet

loneil:
To handle the new `didcomm://?_url=<base64 encoded url>` type handler described in https://github.com/bcgov/vc-authn-oidc/issues/504

The deep link will be formatted as
`bcwallet://aries_proof-request?_url=<vcauth payload redirect URL>`

The is in an upcoming BC Wallet release so I've added a feature toggle as a `useUrlDeepLink` helm value so it can switch back and forth between the formats to accommodate app release or if there's issues with the app linking.

Usage is at https://github.com/bcgov/vc-authn-oidc/pull/557/files#diff-5207e0c2282ef37eda5e5370ef903ef9343b7fb1e37e61ae7192a3ef55693289R187

Also update dependencies


loneil:
Handling here https://github.com/bcgov/vc-authn-oidc/pull/557
BC Wallet release with this is slated to be out next Monday, probably still worth having the config toggle in there though IMO for any needs to troubleshoot it back and forth to the older format or anything for a while.

esune (approved):
Changes look good 👍🏻

Issue #558, opened 2024-06-19
The created_at auth session timestamps are static


loneil:
Seeing on dev the new `created_at` field in the `auth_session` collection is not actually the record Created At time. Seems to be app startup time

![image](https://github.com/bcgov/vc-authn-oidc/assets/17445138/178f5962-1429-4c0e-b00e-e6a4d119efef)

PR #559, opened 2024-06-19
Fix auth session Created At timestamp

loneil:
See https://github.com/bcgov/vc-authn-oidc/issues/558

Use a factory fxn for default like other created at fields in the controller do.


Gavinok (approved):
Lgtm

PR #560, opened 2024-06-20
Present proof 2

Gavinok:


loneil:
See changes needed for ACA-Py 1.0.0 here. Updated VCauth Python version to match, though in this case they could mismatch if something is wrong with updating (not a plugin situation like Traction)
https://github.com/bcgov/vc-authn-oidc/pull/584

Haven't tested fully everything yet but could experiment with that to see if proofs 2.0 are fully working with that.

I'm not sure we'd want to merge on an RC aca-py though? If a VCauth production bugfix were needed then we'd be sending the RC to production? Maybe it's ok given the limited use cases in VCAuth. Could check with @esune

esune:
> See changes needed for ACA-Py 1.0.0 here. Updated VCauth Python version to match, though in this case they could mismatch if something is wrong with updating (not a plugin situation like Traction) #584
> 
> Haven't tested fully everything yet but could experiment with that to see if proofs 2.0 are fully working with that.
> 
> I'm not sure we'd want to merge on an RC aca-py though? If a VCauth production bugfix were needed then we'd be sending the RC to production? Maybe it's ok given the limited use cases in VCAuth. Could check with @esune

I think we can use the RC temporarily, if needed - chances are we'll be switching to the final release pretty soon though.

Gavinok:
I have updated to now use the new RC and from my limited testing things seem to be working

oidc-controller/api/core/acapy/client.py
```diff
@@ -134,12 +118,13 @@ def oob_create_invitation(
...
             "attachments": [
                 {
-                    "id": presentation_exchange["presentation_exchange_id"],
+                    "id": presentation_exchange["pres_ex_id"],
                     "type": "present-proof",
                     "data": {"json": presentation_exchange},
                 }
             ],
             "use_public_did": use_public_did,
+            "trace": True,
```
esune, line 127: Is this flag required? I think we only want to add trace when troubleshooting, so it should not be set to `true` by default. Might want to have this removed, or follow the `DEBUGGER` environment value maybe.

oidc-controller/api/core/acapy/tests/test_client.py
```diff
@@ -144,48 +144,6 @@ async def test_get_presentation_throws_assertion_error_for_non_200_response_from
...
-    )
-
-    client = AcapyClient()
-    client.agent_config.get_headers = mock.MagicMock(return_value={"x-api-key": ""})
-    try:
-        client.verify_presentation("1234-567890")
-    except AssertionError as e:
-        assert e is not None
-
-
```
esune, line 188: Were these removed because we are using the auto-verify feature of the agent?

esune (changes requested):
Small tweaks requested, overall looking good

oidc-controller/api/core/acapy/tests/test_client.py
Gavinok: Yes, I can port them if we want to keep the functionality but figured it's going to only be dead code as of now.

esune (approved):

esune:
Changes look good to me. Unless @loneil has objections to not merge this based on current wallet issue troubleshooting I think we could move forward.

loneil:
I would hold off while finalizing testing on a few BC Wallet fixes that are best reproduced with current mainline VCauth-N, rather than introduce another variable so far with the new protocol and the aca-py release. I can do preliminary testing on this branch now though.

https://github.com/bcgov/bc-wallet-mobile/issues/2053
https://github.com/bcgov/bc-wallet-mobile/issues/2052
https://github.com/bcgov/bc-wallet-mobile/issues/2072

oidc-controller/api/routers/oidc.py
```diff
@@ -113,6 +113,9 @@ async def get_authorize(request: Request, db: Database = Depends(get_db)):
 
     # Prepeare the presentation request
     use_public_did = not settings.USE_OOB_LOCAL_DID_SERVICE
+    wallet_did = client.get_wallet_did(public=use_public_did)
+
+    byo_attachment = PresentProofv10Attachment.build(pres_exch_dict["pres_request"])
```
loneil, line 118: These were removed here
https://github.com/bcgov/vc-authn-oidc/pull/576/files

Not sure if merge conflict put them back or something

oidc-controller/api/routers/oidc.py
```diff
@@ -123,9 +126,7 @@ async def get_authorize(request: Request, db: Database = Depends(get_db)):
     else:
         wallet_did = client.get_wallet_did(public=use_public_did)
 
-        byo_attachment = PresentProofv10Attachment.build(
-            pres_exch_dict["presentation_request"]
-        )
+        byo_attachment = PresentProofv10Attachment.build(pres_exch_dict["pres_request"])
```
loneil, line 129: This is the difference in the response from created presentation that is being passed to the byo_attachment

image

jamshale (commented):
Looks good to me :+1: Happy to see that unit testing stuff come to use.

loneil (changes requested):
The `c_i` mode (USE_OOB_PRESENT_PROOF = false) isn't working when I try 2.0 locally. The wallet just spins on "making connection"

For 1.0 pres on a working environment, the attachment's base64 encoded data is just the proof body itself:

```
{
  "name": "proof_requested",
  "version": "0.0.1",
  "requested_attributes": {
    "req_attr_0": {
      "names": ["given_names", "family_name", "country"],
      "restrictions": [
        { "schema_name": "Person", "issuer_did": "L6ASjmDDbDH7yPL1t2yFj9" },
        { "schema_name": "Person", "issuer_did": "QEquAHkM35w4XVT3Ku5yat" },
        { "schema_name": "Person", "issuer_did": "M6dhuFj5UwbhWkSLmvYSPc" }
      ],
      "non_revoked": { "from": 1721682437, "to": 1721682437 }
    }
  },
  "requested_predicates": {},
  "nonce": "314541474598582064527017"
}
```

On this branch though I see the attachment containing the whole pres req response (nested down in there in another base 64 is the proof):

image


I quickly tried digging down to that proof and just using it in the attachment, but the wallet still spun at the same place... so that might be a different issue with the BC Wallet and the type...?
Or I did something wrong.

Out of Band 2.0 proof are working for me for scans and deep links on android and iOS in beta and release version of the wallet.
But I think we'll still be needing to support connectionless (@esune ?) for fallbacks if there's OOB issues/maybe other wallet support.

So if that's the case we should

a) confirm attachement format (should it be the same as 1.0)/
b) if we fix our format, and BC Wallet doesn't work, coordinate to find the issue there.

cvarjao:
@loneil , what version of BC Wallet did you test with? version 1.0.18 was release yesterday

loneil:
@cvarjao for this branch testing proof 2.0 I'm using 1.0.18 build 1801 for both Android and iOS (on the beta on Android but not iOS but looks like same release for today)

esune:
@loneil yes, unless we have a good reason to drop conenction-less right away we should maintain support for now. We'll plan to slowly move away from it though.

cvarjao:
When trying to receive an invite from aca-py via Credo I get some error message about [DID Rotate attachment is missing.](https://github.com/openwallet-foundation/credo-ts/blob/b387574e8ba33ee5667783133610201cb39b7301/packages/core/src/modules/connections/DidExchangeProtocol.ts#L527)

cvarjao:
So, what would be the correct API call for using `did:peer:4`?
I've tried using:
```javascript
http.post(`/out-of-band/create-invitation`,{
            "accept":["didcomm/aip1","didcomm/aip2;env=rfc19"],
            "alias":`Faber\`s 😇 - ${new Date().getTime()}`,
            "goal":"",
            "goal_code":"",
            "handshake_protocols":["https://didcomm.org/didexchange/1.0"],
            "my_label":`Faber\`s 😇 - ${new Date().getTime()}`,
            "protocol_version":"1.1",
            "use_public_did":false,
            "use_did_method": "did:peer:4",
        }, {
            params: {
                "auto_accept": true,
                "multi_use": false,
                "create_unique_did": false,
            },
            headers:{
                'Content-Type': 'application/json',
                'Authorization': `Bearer ${config.auth_token}`
            }
        })
```

esune:
> So, what would be the correct API call for using `did:peer:4`? I've tried using:
> 
> ```js
> http.post(`/out-of-band/create-invitation`,{
>             "accept":["didcomm/aip1","didcomm/aip2;env=rfc19"],
>             "alias":`Faber\`s 😇 - ${new Date().getTime()}`,
>             "goal":"",
>             "goal_code":"",
>             "handshake_protocols":["https://didcomm.org/didexchange/1.0"],
>             "my_label":`Faber\`s 😇 - ${new Date().getTime()}`,
>             "protocol_version":"1.1",
>             "use_public_did":false,
>             "use_did_method": "did:peer:4",
>         }, {
>             params: {
>                 "auto_accept": true,
>                 "multi_use": false,
>                 "create_unique_did": false,
>             },
>             headers:{
>                 'Content-Type': 'application/json',
>                 'Authorization': `Bearer ${config.auth_token}`
>             }
>         })
> ```

Might be best to ask in the ACA-Py Discord or repo rather than in this PR, as this is ACA-Py behaviour (not specific to VC-AuthN).

jamshale:
Just FYI, I think that payload looks correct. If it's not working then I'm not sure why.

esune:
I am able to scan the QR code in BC Wallet and respond to the proof-request (which is happening QUITE quickly), but then the page in the demo app does not update to the success state and navigate away.

esune:
> I am able to scan the QR code in BC Wallet and respond to the proof-request (which is happening QUITE quickly), but then the page in the demo app does not update to the success state and navigate away.

Nevermind, the issue was with me using a CANdy prod credential to respond at the proof, and that ledger not being in the agent config. Everything works as expected.

esune (dismissed):
LGTM 👍🏻

loneil (changes requested):
Tested out the config permutations locally, working for me. The connectionless proof is working now after the attachment change
Just minor change now that we can update to the latest RC of 1.0.0

docker/docker-compose.yaml
```diff
@@ -100,7 +100,7 @@ services:
       - vc_auth
 
   aca-py:
-    image: ghcr.io/hyperledger/aries-cloudagent-python:py3.9-0.12.1
+    image: ghcr.io/hyperledger/aries-cloudagent-python:py3.12-1.0.0rc4
```
loneil, line 103: We can update this to rc5 now I believe https://github.com/hyperledger/aries-cloudagent-python/pkgs/container/aries-cloudagent-python/247708442?tag=py3.12-1.0.0rc5

esune (approved):

loneil (approved):

Issue #561, opened 2024-06-21
Persistent subject_identifier is an object


esune:
When the `generate_consistent_identifier` flag to true for a presentation configuration, `vc-authn` canonicalizes the available presented attributes and places them in the `sub` claim of the token. This is not desirable, especially since the subject identifier `MUST NOT exceed 255 ASCII characters in length` (quoting the OpenID Spec. Additionally, a JSON object may include characters that require serialization/escaping.

The generate consistent identifier option should calculate a hash (suggested sha256) of the canonicalized JSON object currently used as a subject identifier to meet the field size restrictions as well as to make the value actually usable.

PR #562, opened 2024-06-21
Use hash for sub

loneil:
See https://github.com/bcgov/vc-authn-oidc/issues/561 for details

Create the canonicaljson bytes the same as before but hash that with sha256 for a fixed 64 chars

At the point where the token is gotten from:
image

Now get a sub that looks like this:
image


esune (approved):
LGTM :+1
We'll likely want to get this scenario covered by unit tests as well in the future.

Issue #537, opened 2024-06-05
Migrate GHA deployment workflow to ArgoCD triggering


i5okie:
Existing GitHub Actions workflow utilizing Helm upgrade command should be migrated to trigger ArgoCD sync.

The workflow will need to update relevant files in the GitOps repository.
Pushing the changes will trigger ArgoCD to synchronize the application.

part of: https://github.com/bcgov/DITP-DevOps/issues/158#issuecomment-2145969763

i5okie:
Completed with this PR: https://github.com/bcgov/vc-authn-oidc/pull/564

PR #564, opened 2024-06-24
Migrate GHA deployment workflow to ArgoCD triggering

i5okie:
### - Update Build & Deploy Development workflow
- Add logic to check for changes to the Helm chart files
- Update `trust-over-ip-configurations` repo with chart/app version and image tag as necessary
- Trigger remote workflow to propagate changes to the GitOps repo
- Trigger ArgoCD Sync if no Helm chart changes, otherwise deploy with the chart changes

### - Update Helm Chart Release workflow
- Update Chart versions for Test and Production environments in the GitOps repo
- Explicitly trigger ArgoCD Sync via webhook

### - Miscellaneous
- Update workflows and actions to use the latest stable publicly available actions

### - Note
When updating the Helm chart default values, **changes must be merged** into `trust-over-ip-configurations` repo **before** merging a PR in this repo.

---

Addresses https://github.com/bcgov/vc-authn-oidc/issues/537
Part of https://github.com/bcgov/DITP-DevOps/issues/158#issuecomment-2145969763


.github/workflows/chart_release.yaml
```diff
@@ -27,3 +27,58 @@ jobs:
...
+          cd ministry-gitops-ditp
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+      - name: Lookup latest chart
+        id: chart_version
+        run: |
+          helm repo add vc-authn-oidc https://bcgov.github.io/vc-authn-oidc
+          helm repo update
+          echo "::set output name=APP_VERSION::$(helm search repo vc-authn-oidc --version ^0.2 -ojson | jq '.[0].app_version')"
+          echo "::set output name=CHART_VERSION::$(helm search repo vc-authn-oidc --version ^0.2 -ojson | jq '.[0].version')"
```
esune, line 62: I sthere a way to specify the version so that we do not need to keep this up-to-date manually? If I am reading this correctly it is looking for versions starting with `0.2` and we could move to `0.3` any time, and this would fail to execute?

.github/workflows/on_push_main.yaml
```diff
@@ -14,14 +14,79 @@ jobs:
...
+          cd ministry-gitops-ditp
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+      - name: Lookup latest chart
+        id: chart_version
+        run: |
+          helm repo add vc-authn-oidc https://bcgov.github.io/vc-authn-oidc
+          helm repo update
+          echo "::set output name=APP_VERSION::$(helm search repo vc-authn-oidc --version ^0.2 -ojson | jq '.[0].app_version')"
+          echo "::set output name=CHART_VERSION::$(helm search repo vc-authn-oidc --version ^0.2 -ojson | jq '.[0].version')"
```
esune, line 63: Same as above

.github/workflows/chart_release.yaml
i5okie: It's a way to show only versions 0.2 or newer. filtering out anything older.
We do not need to keep updating it.
An older chart will not work, as it does not have the required updates to work with Argo pipeline.

esune: Unless we can specify a "greater than" clause I think it might be best to remove, since otherwise we need to maintain the filter in the action if using semantic versioning (for anything other than a patch release)

esune (approved):
LGTM 👍🏻
Might want a peek/confirmation from @WadeBarnes before merging.

.github/workflows/chart_release.yaml
```diff
@@ -28,57 +28,74 @@ jobs:
...
           yq e -i '.version = $CHART_VERSION' services/vc-authn-oidc/charts/prod/Chart.yaml
-      - name: Commit and push changes
+
+      - name: Commit and Push to trust-over-ip-configurations Repo
         run: |
-          cd ministry-gitops-ditp
+          cd trust-over-ip-configurations
           git add services/vc-authn-oidc/charts/test/Chart.yaml services/vc-authn-oidc/charts/prod/Chart.yaml
-          git commit -m "Update chart version"
+          git commit -m "Update chart and app versions" || echo "No changes to commit"
```
esune, line 83: I suspect this will still output the git command result AND add the echo afterwards, is that the intention?

.github/workflows/on_push_main.yaml
```diff
@@ -14,96 +14,131 @@ jobs:
...
+          
+      - name: Commit and Push to trust-over-ip-configurations Repo
         run: |
-          cd ministry-gitops-ditp
+          cd trust-over-ip-configurations
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
           git add services/vc-authn-oidc/charts/dev/values.yaml services/vc-authn-oidc/charts/dev/Chart.yaml
-          git commit -m "Update chart version and image tags"
+          git commit -m "Update chart version and image tag" || echo "No changes to commit"
```
esune, line 87: Same as above

.github/workflows/chart_release.yaml
i5okie: The intention here is if git commit command fails, it will echo the "No changes to commit".

esune (approved):

PR #576, opened 2024-07-08
fix: get wallet did only when not using USE_OOB_PRESENT_PROOF

genaris:
Some minor optimizations for the newly introduced `USE_OOB_PRESENT_PROOF` mode:

- Do not ask ACA-Py for wallet DID in such case
- Don't create any PPV1 attachment at this step, since it will be created by ACA-Py
- Also don't call ACA-Py client constructor twice


loneil (approved):
Good catches, works for me. Thanks

PR #577, opened 2024-07-08
fix: add handshake_protocols to oob invitation

genaris:
I noticed that `handshake_protocols` field is gone when ACA-Py client parses the response. Usually it is not used in BCGov main use-case but I'm trying to use it in a case where I want to use a public DID and re-use any connection with the service.


loneil (approved):
Yes in our current use case we did not need handshake_protocols, I tested it with and without an empty array with this change and still working as expected in our case.

Thanks for the fix!

PR #578, opened 2024-07-09
feat: custom wallet URI scheme

genaris:
A proposal to make wallet deep link URI scheme configurable, so it may not only be used with BC Wallet (bcwallet://) but also with any Aries-capable agent that implements the concept defined in [Aries RFC 0268](https://github.com/hyperledger/aries-rfcs/blob/main/concepts/0268-unified-didcomm-agent-deeplinking/README.md).

It basically consists on using `WALLET_DEEP_LINK_PREFIX` environment variable on the controller to set the URI scheme and host used as a prefix. For example, 'didcomm://oob-invitation' for URLs such as `didcomm://oob-invitation?c_i={base64encoded msg}`.

It keeps the default `bcwallet://aries_proof-request` to make this straightforward for BCGov use case.

I hope you like it as much as we had fun using this project :tada:


loneil (dismissed):
We've discussed making this configurable, hadn't gotten around to adding the config so this is exactly what we'd set up. Thanks! 😄 

I'll note we use `aries_proof-request` after the URI scheme, but it doesn't seem to matter to the BC Wallet what that string is. So that _could_ optionally be configurable I guess if other wallets may care about that or something.

esune (approved):
LGTM 👍🏻 
Thanks @genaris!

PR #582, opened 2024-07-16
Remove handshake protocols again temporarily

loneil:
The OOB proof pydantic model was missing handshake_protocols so that was never getting outputted in the OOB message. Even though aca-py was returning it with `[]` in there.

https://github.com/bcgov/vc-authn-oidc/pull/577 addressed this so it could be used in the event a VCAuth operator does want something in there.

Could have sworn I tried it out a bunch with BC Wallet after that but I must have misconfigured something because now if I turn on OOB I get an error when scanning in BC Wallet
image

If I set didcomm or connections in the handshake protocols then the wallet seems to spin on the connecting screen.
So something is either wrong in the Wallet or our OOB message setup when including handshake protocols and going back to the old situation where it is just missing from the output works.

Trying to troubleshoot other OOB issues in BC Wallet so one thing at a time and going to eliminate this error from the current situation.

### DIFF
--- a/oidc-controller/api/core/aries/out_of_band.py
+++ b/oidc-controller/api/core/aries/out_of_band.py
@@ -26,6 +26,7 @@ class OutOfBandMessage(BaseModel):
         alias="requests~attach"
     )
     services: List[Union[OOBServiceDecorator, str]] = Field(alias="services")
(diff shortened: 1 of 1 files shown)


loneil:
Added https://github.com/bcgov/vc-authn-oidc/issues/583 to really sort this out when there's time.

esune (approved):

PR #585, opened 2024-07-17
"Back to" URL

loneil:
Add a "Back To App" button that will return to a URL specified by the calling application.

image

image

Will need to use a forwarded query param in the IDP in Keycloak for this (redirect URI will not work for this)
image

See notes in:
https://stackoverflow.developer.gov.bc.ca/questions/1240

RC discussion with SSO team
https://chat.developer.gov.bc.ca/channel/sso?msg=nha9LiEGKbWxjrPHj

### DIFF
--- a/demo/vue/app/frontend/src/components/base/BaseAuthButton.vue
+++ b/demo/vue/app/frontend/src/components/base/BaseAuthButton.vue
@@ -30,7 +30,10 @@ export default {
     login() {
       if (this.keycloakReady) {
         window.location.replace(
-          this.createLoginUrl() + '&pres_req_conf_id=' + this.presReqConfId
+          this.createLoginUrl() +
+            '&pres_req_conf_id=' +
+            this.presReqConfId +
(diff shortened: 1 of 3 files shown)


demo/vue/app/frontend/src/components/base/BaseAuthButton.vue
```diff
@@ -30,7 +30,10 @@ export default {
     login() {
       if (this.keycloakReady) {
         window.location.replace(
-          this.createLoginUrl() + '&pres_req_conf_id=' + this.presReqConfId
+          this.createLoginUrl() +
+            '&pres_req_conf_id=' +
+            this.presReqConfId +
+            '&back_to_url=http://localhost:8080/',
```
loneil, line 36: Hardcoded in simple demo app (like where we set the port). This is an example of where a calling app would provide environment-configured URL calling to VCAuth

esune (approved):

PR #598, opened 2024-07-31
Workflow does not detect changes to the helm chart files

i5okie:
Replace `chart-testing` action with git diff.

### DIFF
--- a/.github/workflows/on_push_main.yaml
+++ b/.github/workflows/on_push_main.yaml
@@ -23,16 +23,14 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
-      
-      - name: Set up chart-testing
-        uses: helm/chart-testing-action@v2.6.1
 
-      - name: Run chart-testing (list-changed)
+      - name: Check for chart changes
         id: list-changed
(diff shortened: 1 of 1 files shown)


.github/workflows/on_push_main.yaml
```diff
@@ -23,16 +23,14 @@ jobs:
...
-      - name: Run chart-testing (list-changed)
+      - name: Check for chart changes
         id: list-changed
         run: |
-          changed=$(ct list-changed --target-branch ${{ github.event.repository.default_branch }})
-          if [[ -n "$changed" ]]; then
+          if git diff --name-only ${{ github.event.before }} ${{ github.sha }} | grep -q "^charts/"; then
             echo "changed=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "changed=false" >> "$GITHUB_OUTPUT"
```
esune, line 33: Does this need to match the variable at line 59 for things to work?

i5okie: These two lines are to make the code more readable. Otherwise, reporting false is not necessary.

esune (approved):

PR #606, opened 2024-08-06
Update ACA-Py 1.0.0 RC and other libraries

loneil:
Go from RC5 to RC6 for agent.

Update other Vue and Python dependencies


Gavinok (approved):
Lgtm

PR #617, opened 2024-08-26
Always deploy dev environment from main

esune:
Always deploy `dev` using chart in `main` branch. We will need to disable ArgoCD for `dev` to avoid issues with synchronization.


i5okie (approved):
lgtm

PR #626, opened 2024-09-11
Add variable substitution for proof configs

loneil:
Allow proof configs to have variable values that will be substituted at runtime when the proof is generated.

The way this works is that when the proof is generated from the config (`generate_proof_request`) it calls a helper class function `replace_proof_variables` that recurses over the generated proof.
This finds any nested value that starts with `$` and then looks for if  the `$<var_name>` is contained in a map of preset variables and runs a function that is mapped to that variable.

That recursive helper method invokes another class in `variableSubsitutions.py` that contains a "static variable" map of variable names, and also overrides for `contains` and `getitem` allowing "dynamic variables" that can have parameters in the variable name.

So this allows a VCAuth operator to potentially just drop in their own `variableSubsitutions.py` allowing for operator-defined variable allowances and calculations. 

Supported here are 
`$now` get now time.time()
`$today_str` string rep of today's date ie 20240911
`$tomorrow_str` string rep of tomorrow date ie 20240912

and a dynamic variable `$threshold_years_X` that will do a string rep of today minus X years (like the 19 year check in LCRB DAV)
So adding `$threshold_years_10` would substitute in 20240911

Added unit test coverage for this stuff.

**Error handling**
If a proof has a `$<var_name>` that does not resolve, throw back a 400 when trying to get the proof

image

### DIFF
--- a/oidc-controller/api/verificationConfigs/helpers.py
+++ b/oidc-controller/api/verificationConfigs/helpers.py
@@ -1,6 +1,14 @@
 from .variableSubstitutions import variable_substitution_map
 
 
+class VariableSubstitutionError(Exception):
+    """Custom exception for if a variable is used that does not exist."""
+
+    def __init__(self, variable_name: str):
+        self.variable_name = variable_name
+        super().__init__(f"Variable '{variable_name}' not found in substitution map.")
+
+
 def replace_proof_variables(proof_req_dict: dict) -> dict:
     """
     Recursively replaces variables in the proof request with actual values.
@@ -25,8 +33,11 @@ def replace_proof_variables(proof_req_dict: dict) -> dict:
                 if isinstance(i, dict):
                     replace_proof_variables(i)
         # If the value is a string and matches a key in the map, replace it
-        elif isinstance(v, str):
+        elif isinstance(v, str) and v.startswith("$"):
             if v in variable_substitution_map:
                 proof_req_dict[k] = variable_substitution_map[v]()
+            else:
+                raise VariableSubstitutionError(v)
+
         # Base case: If the value is not a dict, list, or matching string, do nothing
     return proof_req_dict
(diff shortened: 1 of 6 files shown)


jamshale (approved):
I'm not very familiar with this codebase anymore but the code looks really good. I like the unit testing with the feature.

I'm not too sure about the `$` placeholder. I think it's really unlikely it gets used and matches the substitution map. If you wanted to be more careful you could possibly think of a placeholder even less likely to be used that has multiple characters.

I will approve, but other contributors should have a look.

oidc-controller/api/verificationConfigs/helpers.py
```diff
@@ -0,0 +1,43 @@
...
+
+    Returns:
+        dict: The updated proof request dictionary with placeholder variables replaced.
+    """
+
+    for k, v in proof_req_dict.items():
+        # If the value is a dictionary, recurse
+        if isinstance(v, dict):
+            replace_proof_variables(v)
+        # If the value is a list, iterate trhough list items and recurse
```
jamshale, line 30: small spelling mistake here.

esune (approved):
Looks good! I would not worry too much about the `$` being a possible "real" value, I think we just need to outline what the templating language expects and its limitations and go with it. In the end, we also will be approving/rejecting (and probably helping compile) the proof-requests that will make it into VC-AuthN.

I don't see a reason to process anything other than the `predicates` section, at least not now. If the processing time isn't significantly higher we may leave the whole payload for processing though.

loneil:
> I don't see a reason to process anything other than the `predicates` section, at least not now. If the processing time isn't significantly higher we may leave the whole payload for processing though.

Yeah it goes through the whole payload. It's recursive, but the payload is always going to be such a relatively small bit of JSON in the grand scheme of thigs it won't make any real difference for performance if it was restricted to just the predicates.

oidc-controller/api/verificationConfigs/helpers.py
loneil: Thx, will throw a fix for that in a subsequent dependency update to not have to re-request review here 😄

PR #627, opened 2024-09-12
Configure invitation label for proofs

loneil:
The `my_label` in the OOB invitation that we use to contain the proof can be set to control what the BC Wallet displays as the requestor.

image

Set environment and helm config to allow this to be set in an environment. If not provided default is just "VC-AuthN".

We _could_ set this dynamically per-proof, allowing control at the presentation config level, but discussing we're just going to go as an environmental setting for now. Easy to allow overrides in the granular configs at some point if that changes.

Dependency updates also


loneil:
Pre-setting candidates for values here https://github.com/bcgov/trust-over-ip-configurations/pull/237

WadeBarnes (approved):

esune (approved):

PR #640, opened 2024-09-24
add helm chart entry for managing the ttl index

### LINKED ISSUE #539: Update Helm chart to support TTL index creation
#531 introduced a mechanism to set-up an optional TTL index in mongo that purges old records after they "expire". The default behaviour is to not have a TTL set-up, but we will want to have these settings in the Helm chart for ease of deployment.

Gavinok:
This PR resolves #539


i5okie (approved):
looks good to me

PR #641, opened 2024-09-27
Documentation for variable subs

loneil:
(Note: do not need to include in release we just made)

Document proof request variables in VCAuthN. Added a _Proof Substitution Variables_ section to `ConfigurationGuide.md` and linked to it from a couple other spots.

Also some library updates

### DIFF
--- a/docs/ConfigurationGuide.md
+++ b/docs/ConfigurationGuide.md
@@ -23,7 +23,7 @@ To input settings manually, or review them:
 
 - **Token URL**: this must be set to `{PUBLIC_VC_AUTHN_URL}/token`
 
-- **DIsable User Info**: it is recommended to disable the user info endpoint, since VC Authn does not store/provide user information.
+- **Disable User Info**: it is recommended to disable the user info endpoint, since VC Authn does not store/provide user information.
 
 - **Client ID/Client Secret**: these settings will be used to identify and secure the IdP integration between Keycloak and VC Authn. Make sure the **client secret** parameter is unique to your VC Authn instance. VC-AuthN supports two methods of client authentication: `Client secret sent as basic auth` and `Client secret sent as post`.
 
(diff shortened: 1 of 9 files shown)


docs/ConfigurationGuide.md
```diff
@@ -90,3 +90,47 @@ Several functions in VC-AuthN can be tweaked by using the following environment
...
+See the `oidc-controller\api\verificationConfigs\variableSubstitutions.py` file for implementations.
+
+#### Customizing variables
+
+In `oidc-controller\api\verificationConfigs\variableSubstitutions.py` there are the built-in variables above.
+For an advanced use case, if you require further customization, it could be possible to just replace that `variableSubstitutions.py` file 
+in a VC-AuthN implementation and the newly introduced variables would be run if they are included in a proof request configuration.
+
+For regular variables they can be added to the `static_map`, mapping your variable name to a function doing the operation.  
+For "dynamic" ones alter `__contains__` and `__getitem__` to use a regex to parse and extract what is needed.
```
esune, line 134: This got me thinking: what about being able to specify a file that `adds` or `replaces` variable substitutions to the `static_map`, rather than replacing the file?

i.e.: default variable substitutions are always available, but if a separate file path is specified we take it's function members and add them to the map so they can be used as well. Not for this PR, but dropping the comment here as it is in context.

esune (approved):
Docs look good, thanks! 👍🏻

docs/ConfigurationGuide.md
loneil: Yeah we could probably have it look for the presence of another file and then import functions. The current model would need an operator to overwrite the existing file (and could end up with conflicts if the default file gets changed etc).

esune: I'll log something for us to look into - that was exactly my thinking.

Issue #583, opened 2024-07-16
Sort out OOB Handshake Protocols


loneil:
OOB proofs are working to the BC Wallet provided we do not put anything in `handshake_protocols`

The call to ACA-Py to make the OOB includes the handshake_protocols in the response, but we were not including that in our model (by mistake!) so it does not end up in the QR Code/Deep Link payload.
Fixing that and having it go through as `[]` causes an invalid QR code with an error in the wallet

![image](https://github.com/user-attachments/assets/2e5b1339-971e-40c6-9234-1255fff5d227)

Setting content in there, didcomm or connections causes the wallet to spin.

Some context here: https://github.com/bcgov/vc-authn-oidc/pull/582

I don't know if we are missing some other fields in our OOB invitation creation in order to be using handshake_protocols as a blank, or filled array (or which of those cases we'd want for sure)
Or not sure if this is an issue on the BC Wallet side, or Credo or something...
need some investigation then can add back in handshake_protocols to our model.

esune:
Might be related to this: https://github.com/openwallet-foundation/credo-ts/blob/808878decbe66909fa9568284ae10cf29201f50d/packages/core/src/modules/oob/OutOfBandApi.ts#L423

cvarjao:
The error message is related to:
https://github.com/search?q=repo%3Aopenwallet-foundation%2Fcredo-ts+%22Handshake+protocols%22&type=code

esune:
> The error message is raised from: https://github.com/search?q=repo%3Aopenwallet-foundation%2Fcredo-ts+%22Handshake+protocols%22&type=code

Thanks @cvarjao. Do you think we could expect an empty array to be handled/treated as if the option was not set (this is what ACA-Py seems to do), or do we need to gather consensus on what "optional" attribute means first? Trying to figure out where to route this issue next, since VC-AuthN is just a "consumer".

cvarjao:
I do not understand the protocol at the level to provide any meaning feedback, but sounds like credo/aca-py need to have some alignment on how it is implemented though.

wadeking98:
I'm running into this after my fix for the `_url` deeplink issue (https://github.com/bcgov/bc-wallet-mobile/issues/2053). I can fix the original error but after that I run into this error because the handshake_protocols is just an empty array in the message body

esune:
[...]

PR #650, opened 2024-10-03
Add back handshake protocols

loneil:
Add back in the handshake protocols field in the model for OOB invitations. See details here https://github.com/bcgov/vc-authn-oidc/issues/583

BC Wallet has released a while back with the Credo version that handles this now.

We provide a blank array in default setup for VCAuth at this time:

image


esune (approved):

loneil:
> LGTM 👍🏻
> 
> @loneil you were able to test this so we can merge, right?

Yeah can go ahead, tested out on iPad release version and Android beta

PR #656, opened 2024-10-08
Append @pres_req_conf_id to sub

### LINKED ISSUE #538: Improvements to subject identifier generation
In order to support scenarios where multiple `pres_req_conf_id` could yield the same value for the subject identifier (i.e.: a field with the same value is chosen as sub, the generated sub based on the credential payload is the same because of the content of the credential).

In order to further "namespace" authentication requests, the subject identifier should be built with `@pres_req_conf_id` as a suffix, so that each proof-configuration will return a different value for `sub` regardless of whether the credential contents are the same.

Gavinok:
This PR resolves #538 

This resolves overlapping subject identifiers by appending @pres_req_conf_id when generating the sub.

### DIFF
--- a/oidc-controller/api/core/oidc/issue_token_service.py
+++ b/oidc-controller/api/core/oidc/issue_token_service.py
@@ -110,8 +110,13 @@ class Token(BaseModel):
             # Do not create a sub based on the proof claims if the
             # user requests a generated identifier
             # Generate a SHA256 hash of the canonicaljson encoded proof_claims
-            encoded_json = canonicaljson.encode_canonical_json(proof_claims)
-            sha256_hash = hashlib.sha256(encoded_json).hexdigest()
+            encoded_json: bytes = canonicaljson.encode_canonical_json(proof_claims)
+            sha256_hash = hashlib.sha256(
+                (
+                    encoded_json
+                    + f"@{auth_session.request_parameters['pres_req_conf_id']}".encode()
+                )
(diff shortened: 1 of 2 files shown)


oidc-controller/api/core/oidc/issue_token_service.py
```diff
@@ -98,7 +98,14 @@ def get_claims(
...
-            oidc_claims.append(Claim(type="sub", value=sub_id_claim.value))
+            assert type(auth_session.request_parameters["pres_req_conf_id"]) == str
+            oidc_claims.append(
+                Claim(
+                    type="sub",
+                    value=f"{sub_id_claim.value}@{auth_session.request_parameters['pres_req_conf_id']}",
+                )
+            )
+
         elif ver_config.generate_consistent_identifier:
```
esune, line 109: You're going to want to append value of the `pres_req_conf_id` also when generating a consistend `sub` as the generated value will depend on the proof-request configuration.

In that case, you might want to save the value of `pres_req_conf_id` before the `if/else` so you can use it in both places without having to explicitly access the attribute in two spots.

esune (changes requested):

oidc-controller/api/core/oidc/issue_token_service.py
Gavinok: If I understand this correctly this issue should now be resolved

esune (approved):

PR #657, opened 2024-10-09
Helm Chart Improvements

i5okie:
- Do not create the database secret, use the secret created by PostgreSQL chart itself when enabled.
- Update Values file
- Update _helpers
- Update and clean up Readme

**Note:**
_When installing with a custom release name that does not contain "vc-authn-oidc", the user must set `fullNameOverride`._


esune (approved):
Changes look good. Are you going to coordinate the (manual) updates that need to happen in `dev` when switching to this release (i.e.: as soon as it is merged to main) as well as for `test`/`prod` in the near future?

Issue #544, opened 2024-06-12
Pydanctic .dict() depcrecation warnings


loneil:
Places using the Pydantic models and `.dict()` show a warning

![image](https://github.com/bcgov/vc-authn-oidc/assets/17445138/84e15dbd-5f79-40ee-a651-87d5ed6d9921)
 
Can(?) I think just replace like:
```
response = client.create_presentation_request(ver_config.generate_proof_request())
pres_exch_dict = response.model_dump()
```

esune:
Resolved in #658

PR #658, opened 2024-10-11
Python version upgrade and remove deprecated functions

### LINKED ISSUE #605: Python 3.12
OIDC controller is presently running python 3.11.
Should update and test out 3.12 (the ACA-Py agent is using 3.12 as part of being on ACA-Py 1.0.0)

Gavinok:
This PR resolves #605 as well as #544 

In addition to upgrading the python version I also have removed multiple deprecation notices related to types, the `dict()` method, `parse_obj()` method, and updated the pytest settings.


esune (approved):
👍🏻

PR #675, opened 2024-10-30
Add User Definable Variable Substitution

### LINKED ISSUE #649: Custom variable substitutions from external file
Rather than replacing the default file containing the variable substitution definitions/implementations, it would be better to allow users to:
- specify a path to a python file containing custom definitions (or overrides) for the variable substitutions
- if set, try to load the new definitions and add them to the map of existing substitutions
  - nice to have: log which custom substitutions were loaded
- if set, but file does not exist/is unreachable, log a warning and carry on

Gavinok:
This PR resolves #649

I have documented and added a user definable way to add new variable substitutions. A simple API for defining regex based variable matching. Here is a simple example usage.

```python
def sub_days_plus_one(days: str) -> int:
    """Strings like '$sub_days_plus_one_4' will be replaced with the
    final number incremented by one. In this case 5.
    $sub_days_plus_one_4 -> 5
    $sub_days_plus_one_10 -> 10"""
    return int(days) + 1


variable_substitution_map.add_variable_substitution(
    r"$sub_days_plus_one_(\d+)", sub_days_plus_one
)
```


esune (commented):
I do like how this is turning out. We need to add the override file bits to the Helm chart as well, maybe with a basic empty file definition that could be customized?

loneil:
Think this works for me, good idea to have sample functions in there that would be in effect for a deployment. Yeah same question about what helm setup would be

Gavinok:
I am going with Emilano's idea for now. I will setup the project with an empty python file that will be loaded allowing the users to extend it as they please

charts/vc-authn-oidc/templates/deployment.yaml
```diff
@@ -133,6 +141,9 @@ spec:
             - name: auth-session-ttl
               mountPath: /home/aries/sessiontimeout.json
               subPath: sessiontimeout.json
+            - name: custom-variable-substitution
+              mountPath: /home/aries/user_variable_substitution.py
+              subPath: sessiontimeout.json
```
esune, line 146: This should be `user_variable_substitution.py`

docs/ConfigurationGuide.md
```diff
@@ -125,12 +125,33 @@ and at runtime when the user navigates to the QR code page, the proof would incl
...
+variable to point at a python file defining new substitutions.
 
-In `oidc-controller\api\verificationConfigs\variableSubstitutions.py` there are the built-in variables above.
-For an advanced use case, if you require further customization, it could be possible to just replace that `variableSubstitutions.py` file 
-in a VC-AuthN implementation and the newly introduced variables would be run if they are included in a proof request configuration.
+##### User Defined Variable API
+In `oidc-controller\api\verificationConfigs\variableSubstitutions.py`
+you will find the method `add_variable_substitution` which can be used
+to modify the existing instance of `VariableSubstitutionMap` named
+`variable_substitution_map`.
```
esune, line 135: Wondering if we need to explicitly call-out how a substitution is loaded, if all a user need to do is provide a well-structured file with new substitutions?

esune (changes requested):

esune (approved):
I think this works, thanks @Gavinok 

@loneil if you want to give it a review in case I missed anything I think we could merge after that.

loneil (approved):
I _think_ this will work deployment-wise. Probably just put it through it's paces on a dev deploy and make sure

PR #678, opened 2024-11-08
Human readable QR

### LINKED ISSUE #646: Add human-readable proof-request information to QR code page
When presented with the QR code to scan with BC Wallet, users will NOT know until after scanning it what is requested in the proof.

To improve this, we should add a `metadata` object to the proof-configuration that can be used to hold information about the proof (useful also for searching in a list) with information about what is requested.

The page would display a section similar to the following (example for the Lawyer + Person credential proof):
The proof request will ask you to prove the following:
- That you are a BC Resident
- That you are a member of the Law Society of British Columbia

The data structure containing the information to be displayed could be something like:
```json
{
 "ver_config_id":"accredited-lawyer-bcpc",
  ...,
  "metadata": {
    "proved_claims": ["That you are a BC Resident", "That you are a member of the Law Society of British Columbia"]
  }
}
```

Note: `i18n` descriptions could be added to accomodate for multi-language systems.

Gavinok:
This pr resolves #646 

The key changes is the addition of the new metadata entry in the VerificationConfig that can contain the title of the verification and the claims it verifies. Here an example of the UI change on mobile.

and for desktop

### DIFF
--- a/oidc-controller/api/core/oidc/tests/__mocks__.py
+++ b/oidc-controller/api/core/oidc/tests/__mocks__.py
@@ -1341,6 +1341,13 @@ auth_session = AuthSession(
 ver_config = VerificationConfig(
     ver_config_id="verified-email",
     subject_identifier="email",
+    metadata={
+        "title": "Get Name",
+        "claims": [
+            "That you are a BC Resident",
+            "That you are a member of the Law Society of British Columbia",
+        ],
+    },
     proof_request=VerificationProofRequest(
         name="BCGov Verified Email",
         version="1.0",
--- a/oidc-controller/api/routers/oidc.py
+++ b/oidc-controller/api/routers/oidc.py
@@ -212,6 +212,12 @@ async def get_authorize(request: Request, db: Database = Depends(get_db)):
         "controller_host": controller_host,
         "challenge_poll_uri": ChallengePollUri,
         "wallet_deep_link": wallet_deep_link,
+        "title": (
+            ver_config.metadata.title
+            if ver_config.metadata
+            else "Scan with a Digital Wallet"
+        ),
+        "claims": (ver_config.metadata.claims if ver_config.metadata else []),
     }
 
     # Prepare the template
--- a/oidc-controller/api/verificationConfigs/models.py
+++ b/oidc-controller/api/verificationConfigs/models.py
@@ -1,4 +1,5 @@
 import time
+from typing_extensions import TypedDict
 from pydantic import BaseModel, ConfigDict, Field
 
 from .examples import ex_ver_config
@@ -38,11 +39,17 @@ class VerificationProofRequest(BaseModel):
     requested_predicates: list[ReqPred]
 
 
+class MetaData(BaseModel):
+    title: str = Field()
+    claims: list[str] = Field()
+
+
 class VerificationConfigBase(BaseModel):
     subject_identifier: str = Field()
     proof_request: VerificationProofRequest = Field()
     generate_consistent_identifier: bool | None = Field(default=False)
     include_v1_attributes: bool | None = Field(default=False)
+    metadata: MetaData | None = Field(default=None)
 
     def get_now(self) -> int:
         return int(time.time())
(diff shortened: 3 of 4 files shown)


esune (commented):
I think this should work well!

The only thing is that - at least to me - the alignment of the new paragraph seems to be off when compared with the rest of the page (which seems to be more centered). Is it true or are my eyes tricking me into seeing that?

oidc-controller/api/verificationConfigs/models.py
```diff
@@ -38,11 +39,17 @@ class VerificationProofRequest(BaseModel):
     requested_predicates: list[ReqPred]
 
 
+class MetaData(BaseModel):
+    title: str = Field()
+    claims: list[str] = Field()
+
```
loneil, line 45: Think we'd set claims to be allowable to be missing or empty (in the event there's only a title)
I think I'm getting an error in one of those cases?

image

```
1 validation error for VerificationConfig\nmetadata.claims\n  Field required [type=missing, input_value={'title': '123'}, input_type=dict]\n    For further information visit https://errors.pydantic.dev/2.9/v/missing\n", "logger": "api.main", "level": "error", "timestamp": "2024-11-08T23:38:01.766230Z"}
```

Empty array seems all good:
image

loneil (changes requested):

loneil:
> I think this should work well!
> 
> The only thing is that - at least to me - the alignment of the new paragraph seems to be off when compared with the rest of the page (which seems to be more centered). Is it true or are my eyes tricking me into seeing that?

This is the alignment, so paragraph left justified. Think center justifying it would then have multiple list items looking weird, so if that were the case think it would be centering just the paragraph (but then having to offset each list item so they line up)

image

oidc-controller/api/verificationConfigs/models.py
Gavinok: Added support for this thanks for the suggestion

loneil (dismissed):
Think this should do the trick 👍

esune (dismissed):

esune (dismissed):
I think this will work 👍🏻

loneil:
Are we just adding a language array to the claims and not the title?

esune:
> Are we just adding a language array to the claims and not the title?

I somehow missed that, good catch. We should add the same language mapping to the title - maybe we should have the language map as the root element with language and title both nested within it @Gavinok

Gavinok:
Added support for titles thanks for pointing that out. Should have just added that to begin with.

oidc-controller/api/core/oidc/tests/__mocks__.py
```diff
@@ -1341,6 +1341,15 @@
...
     subject_identifier="email",
+    metadata={
+        "title": {"en": "Get Name"},
+        "claims": {
+            "en": [
+                "That you are a BC Resident",
+                "That you are a member of the Law Society of British Columbia",
+            ]
+        },
+    },
```
esune, line 1352: What about using a single language key for each entry, rather than duplicating it?
```suggestion
    metadata={
        "en":{
          "title": "Get Name",
          "claims": [
                "That you are a BC Resident",
                "That you are a member of the Law Society of British Columbia",
          ]
        }
    },
```

oidc-controller/api/routers/oidc.py
```diff
@@ -212,6 +212,20 @@ async def get_authorize(request: Request, db: Database = Depends(get_db)):
         "controller_host": controller_host,
         "challenge_poll_uri": ChallengePollUri,
         "wallet_deep_link": wallet_deep_link,
+        "title": (
+            ver_config.metadata.title.get("en")
```
esune, line 216: It might be best to use a parameter for the language, so that we don;t have to update this condition when parsing a language dynamically

Gavinok: Like a parameter for the function or for the template?

esune: I think what you did will work, once we have multilanguage support in the UI we just need to update the code to grab that value rathe rthan the hardcoded `en` that now only appears in one spot

loneil (approved):
Looks good, tried a couple permutations out locally

esune (approved):

PR #685, opened 2024-11-20
Reorg config files

loneil:
Fix (hopefully) issue with VCAuthN deployment from missing config map. Reorg things so we're just using a single config map instead of intending on one per configuration file. Renamed it "controller-config" instead of specific to "session timeout". Add the "blank" (comments, see below) placeholder user variable sub file to the config map.

Default population for this user variable py file comes from values. So keep the pattern of being able to override it with a values yaml, or operator can have whatever means up updating configmap, or the volume itself, as works for their CICD.
App continues to use environment variable to know which path to look for these files, so can override that as well.

Discussed moving location of files to /app but thinking about that didn't want to volume mount into the directory the code goes into, as that could cause confusion (and if you can alter /app it's like changing the source code anyway instead of "external" configuration), and to keep /app as deployed, built source alone.

Used `/etc/controller-config` instead of `/home/aries`
Don't really care about /etc vs /home but googled practices for config file stuff

> **Best Practices:**
> **System-Wide Configuration**:
> 
> Use the /etc directory for configurations that need to be available to all users or affect system-wide services.
> Keep it organized and maintain proper permissions to ensure system security.
> 
> **User-Specific Configuration**:
> 
> Use the /home directory for configurations that are specific to individual users.
> Store user preferences and application settings in the user’s home directory to avoid conflicts and maintain privacy.

Since this is a system service thing and not a user preference...🤷 again not really important. Important part is to not use "aries" naming any more if we can avoid it. If there's any weird permission thing, or if people think `/home` is a more obvious place to look can change it.

Tested helm install locally on minikube and things seem to be where they're expected (below). But will need to make sure OCP deployment is all good and the Python app doesn't have issues loading.
If we merge this I'll follow through on Dev env and make sure things are deploying ok and the app is picking up the values as expected.

New config map, contains both default file values:
image

Checking on controller pod:
image

### DIFF
--- a/charts/vc-authn-oidc/values.yaml
+++ b/charts/vc-authn-oidc/values.yaml
@@ -40,6 +40,7 @@ walletDeepLinkPrefix: bcwallet://aries_proof-request
 ## @param controller.presentationExpireTime The number of time in seconds a proof request will be valid for
 ## @param controller.sessionTimeout.duration The number of seconds an auth_sessions in the states defined in controllerSessionTimeoutConfig is kept for
 ## @param controller.sessionTimeout.config The json list of auth session states that are safe for deletion
+## @param controller.uservariablesubsitution The placeholder for the user variable substitution file that can be used to add substitution variables
 controller:
   cameraRedirectUrl: wallet_howto
   presentationExpireTime: 300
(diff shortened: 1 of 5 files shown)


charts/vc-authn-oidc/values.yaml
```diff
@@ -49,6 +50,9 @@ controller:
       - expired
       - failed
       - abandoned
+  userVariableSubsitution: |- 
+    # This is a default placeholder Python file 
+    # Add your default content here
```
loneil, line 55: Default "blank" file here

i5okie: Looks good. Do we want to have a commented-out example of a substitution?

loneil: Oh yeah, there's the example file in the source here
https://github.com/bcgov/vc-authn-oidc/blob/main/docker/oidc-controller/config/user_variable_substitution_example.py
so could add a function from that commented out, or just a comment saying they could look at that file

i5okie (approved):
lgtm

charts/vc-authn-oidc/values.yaml
esune: My vote goes to adding a function/example for that file commented-out to make it foolproof.

esune (commented):
Looks good, I think if we add the default example code to the values file for the variable substitution we can merge and test in OCP

Gavinok (approved):

charts/vc-authn-oidc/values.yaml
loneil: Added commented out file

Gavinok (approved):

PR #698, opened 2024-12-21
Library updates

loneil:
Python app and vue demo library updates


pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [tool.poetry]
-name = "acapy-vc-authn-oidc"
+name = "vc-authn-oidc"
```
esune, line 2: Is the name change necessary? Maybe related to https://github.com/openwallet-foundation/acapy-vc-authn-oidc/issues/701 ?

loneil: Whoops no that was a merge mistake. Reversing. Thanks for catching

esune (approved):
LGTM 👍🏻

PR #707, opened 2025-01-17
Helm: remove default resource limits, dependency updates

esune:
Removed default resource limits from helm chart, updated Helm dependencies.

c.c.: @i5okie


i5okie (approved):

charts/vc-authn-oidc/Chart.yaml
```diff
@@ -17,10 +17,10 @@ appVersion: "2.2.2"
 # Charts the vc-authn-oidc service depends on
 dependencies:
   - name: mongodb
-    version: 15.4.3
```
loneil, line 20: Mongo 7 to 8 worked locally with docker

esune:
Reverted the postgres dependency update as it looks like it might require extra manual steps to migrate from 15 to 17.
I also tweaked the docker-compose to make sure versions are consistent with the helm chart (thanks @loneil for the suggestion).

loneil (approved):

PR #714, opened 2025-02-06
[Helm] Update dependency: PostgreSQL 14 to PostgreSQL 16

i5okie:
- Update PostgreSQL dependency to use PostgreSQL version 16 **Breaking Change**
- Update README with updated installation instructions
- Bump chart version

_Upgrading to this chart requires manually upgrading the database files._


esune (approved):
Looks good!

PR #715, opened 2025-02-06
User defined templates

### LINKED ISSUE #713: Allow customization of QR code page
In order to support custom UX for the QR code page, it needs to be possible to replace the default page and assets.

There are two options to do this:
* (Preferred) Adding a volume with the appropriate resources (HTML, CSS, JS) to the container and having a configuration parameter tell the app where to find the resource files.
* Allow for resources to be replaced in the container with a custom (downstream) build step, ensuring that the expected resources are available in the target directory.

Custom resources for the QR page should always include an HTML page, with CSS and JS resources either defined inline or referenced. This should allow support for more complex SPAs that can be built separately and then injected into the deployment.

Gavinok:
This PR attempts to resolve #713 

There are now two supported options. 

1. Add overriding files to the [ConfigMap](charts/vc-authn-oidc/templates/configmap.yaml) for the deployment and the other requires users to 

2. Add additional build step in order to add their own html template directory to the controller image and change the environment variable `CONTROLLER_TEMPLATE_DIR` to point to their new template directory.

**Why 2 Options**
While option 1 prevents the need for additional build steps it is limited by the maximum size of a ConfigMap (1MB). Option 2 allows for more complex user made html templates but requires an additional build step.


docs/ConfigurationGuide.md
```diff
@@ -164,3 +164,46 @@ After loading the python file during the service startup each new user
...
+
+# User Customized QR Page
+
+## ConfigMap Based Overrides
+
+By default the QR page shown to users uses OpenWallet branding users
+are capable of overriding any of these files using the
+`{{ include "global.fullname" . }}-template-overrides`
+[ConfigMap](../charts/vc-authn-oidc/templates/configmap.yaml) in the
+in the helm charts. These files will override the existing files seen
```
loneil, line 176: "in the in the"

docs/ConfigurationGuide.md
```diff
@@ -164,3 +164,46 @@ After loading the python file during the service startup each new user
...
+in [html-templates](../html-templates/).
+
+## Complete Directory Overrides
+
+In addition to allow for more flexibility users are welcome to build
+an image based on `ghcr.io/bcgov/vc-authn-oidc` mounting their a
+complete directory with new assets.
```
loneil, line 183: "their a complete directory"

loneil (commented):
Wondering what people think about default location?
For the override variables it's `CONTROLLER_VARIABLE_SUBSTITUTION_OVERRIDE=/etc/controller-config/user_variable_substitution.py` right? Good to have templates in the same spot rather than `/tmp`? Maybe we add more config files later on so could be good to have things together...

I haven't tried an override locally in docker (just confirmed it runs all good as-is), but I think the solution should work (would get Ivan to give a look through Helm setup on this PR) if I'm understanding right. So anything I put in `/tmp/templates` (or whatever we specify that to be, or the user species with `CONTROLLER_TEMPLATE_DIR`), that should get resolved at runtime and pulled in right?
So it's not necessary to build the image with custom templates? Like if someone adds a volume mount at their template directory could those files be independently controlled as deployment time?

Ideally we (and others) wouldn't be building custom I think, just using the vcauth image and deploying files and setting the path? If I'm reading the config right.

docs/ConfigurationGuide.md
esune: The container registry address should be `ghcr.io/openwallet-foundation/acapy-vc-authn-oidc`

esune (changes requested):
I do like @loneil suggestion to keep all templates together in `/etc/controller-config/` as default directory. We can still provide a way to specify a different location, but that path would make sense in most scenarios.

I am also wondering if we should put a "toggle" for the template configmap to be mounted or not (it would always be created in k8s): if we don't do this, even when using a custom build image we would get the configmap mounted at the target location overriding the modified files.

esune:
> Resolved the grammar and corrected the directory + registry address. Do we also want to change the chart to point at that registry?



> Resolved the grammar and corrected the directory + registry address. Do we also want to change the chart to point at that registry?

Yes please, that was something I must have missed when prepping for the move

oidc-controller/api/core/config.py
```diff
@@ -233,6 +233,9 @@ class GlobalConfig(BaseSettings):
     CONTROLLER_VARIABLE_SUBSTITUTION_OVERRIDE: str | None = os.environ.get(
         "CONTROLLER_VARIABLE_SUBSTITUTION_OVERRIDE"
     )
+    CONTROLLER_TEMPLATE_DIR: str = os.environ.get(
+        "CONTROLLER_TEMPLATE_DIR", "/etc/templates"
```
esune, line 237: Would it make sense to use `/etc/controller-config/templates` to keep all configs within the same directory? Any potential issues when mounting volumes for both the templates AND the index configurations? It might be possible to address this by creating the `/etc/controller-config` folder in the dockerfile with the right permission as an empty container and moving files into it during the build and/or via a mounted volume.

Gavinok: Totally I'll make the change.

esune (dismissed):
Looks good! 👍🏻

charts/vc-authn-oidc/values.yaml
```diff
@@ -64,6 +66,10 @@ controller:
     # variable_substitution_map.add_variable_substitution(
     #     r"\$sub_days_plus_one_(\d+)", sub_days_plus_one
     # )
+  # htmlFileOverrides:
+  #   ver_config_explorer.html: ""
+  #   verified_credentials.html: ""
+  #   wallet_howto.html: ""
```
esune, line 72: After speaking with @i5okie, it sounds like it might be best to provide the users with the ability to specify an existing (manually created) configmap to be mounted with the required contents. This could become something like `customHtmlConfigMapName: "my-custom-name"` and be referenced in the deployment if/when mounting the configmap.

charts/vc-authn-oidc/templates/deployment.yaml
```diff
@@ -39,6 +39,13 @@ spec:
         - name: controller-config
           configMap:
             name: {{ include "global.fullname" . }}-controller-config
+        {{- if .Values.controller.htmlFileOverrides }}
+        - name: controller-templates
+          projected:
+            sources:
+              - configMap:
+                  name: {{ include "global.fullname" . }}-template-overrides
+        {{- end }}
```
esune, line 48: This would be toggled by `.Values.controller.customHtmlConfigMapName`: if a name is specified, the name of the configmap would be `.Values.controller.customHtmlConfigMapName`

charts/vc-authn-oidc/templates/deployment.yaml
```diff
@@ -134,6 +143,10 @@ spec:
               mountPath: /opt/token
             - name: controller-config
               mountPath: /etc/controller-config
+            {{- if .Values.controller.htmlFileOverrides }}
```
esune, line 146: Toggle on `.Values.controller.customHtmlConfigMapName` like above, but definition does not need to change.

charts/vc-authn-oidc/templates/configmap.yaml
```diff
@@ -7,4 +7,15 @@ data:
...
+{{- if .Values.controller.htmlFileOverrides }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "global.fullname" . }}-template-overrides
+data:
+  {{- range $key, $value := .Values.controller.htmlFileOverrides }}
+  {{ $key }}: {{ $value | quote }}
+  {{- end }}
+{{- end }}
```
esune, line 21: Please remove, based on following review comments.

docs/ConfigurationGuide.md
```diff
@@ -164,3 +164,48 @@ After loading the python file during the service startup each new user
...
+# User Customized QR Page
+
+## ConfigMap Based Overrides
+
+By default the QR page shown to users uses OpenWallet branding users
+are capable of overriding any of these files using the
+`controller.htmlFileOverrides` value in
+[values.yaml](charts/vc-authn-oidc/values.yaml).
+These files will override the existing files seen in
+[html-templates](../html-templates/).
```
esune, line 177: Please expand documentation to explain how to use the `customHtmlConfigMapName` to turn on configmap volume substitutions, and what are the expected keys for the configmap to work correctly (i.e.: the file names that will be replaced).

esune (changes requested):
A few more tweaks, sorry for the back-and-forth: after reviewing the current strategy with @i5okie he came up with some extra recommendations that are worth following.

i5okie:
I just created an issue regarding placing files in `/etc` directory. https://github.com/openwallet-foundation/acapy-vc-authn-oidc/issues/721
In the context of this PR, I think placing these templates in `/app/templates` would be more appropriate.

docs/ConfigurationGuide.md
```diff
@@ -170,12 +170,36 @@ indication of this will also be logged
...
+  name: user-template-overrides
+data:
+  verified_credentials.html: "<p>New contents of the verified_credentials.html page</p>"
+  wallet_howto.html: "<p>New contents of the wallet_howto.html page</p>"
+  ver_config_explorer.html: "<p>New contents of the ver_config_explorer.html page</p>"
+```
+
+The name of this new ConfigMap needs to be added to
+`controller.customHtmlConfigMapName` value in
 [values.yaml](charts/vc-authn-oidc/values.yaml).
```
esune:
```suggestion
The name of this new/existing ConfigMap needs to be added to
`controller.customHtmlConfigMapName` value in
[values.yaml](charts/vc-authn-oidc/values.yaml) for it to be mounted.
```

esune (commented):
Looks good, just a couple nitpick suggestions for the docs and we're good to go I think.

esune (approved):

PR #727, opened 2025-02-19
Update QR to pending when phone camera is used to scan QR

### LINKED ISSUE #615: Feedback on the QR page when scanning with device camera (instead of App)
We can now use a device camera with the QR code scan, and that actions a deep link that will invoke the BC Wallet app to open up and go get the proof payload.

This differs from scanning the QR code when already inside the app, which redirects the VCauthN route that delivers the proof.

When using the camera->deep link method, the route that changes the status and sends the webhook message is not invoked. So the page doesn't look like it's pending to the user on the QR screen.

We'll need to address this to get the right UX for camera-scanning.

https://github.com/user-attachments/assets/53ef206b-2e65-4ed6-9225-afdc78b9e530

Gavinok:
This PR resolves #615

https://github.com/user-attachments/assets/44684e58-973a-4ad2-a3c5-e476613ef297


loneil (approved):
Looks good to me, ran locally and tried the 3 flavors of login on Android and iOS

esune (approved):

PR #729, opened 2025-02-21
Migrate from /etc for configuration into /app

### LINKED ISSUE #721: Application configuration files should be moved out of the `/etc` directory.
Application files should be moved out of the `/etc` directory.

https://github.com/openwallet-foundation/acapy-vc-authn-oidc/blob/dc46e70ab9509e92b7036537c0b21ee4b6494640/charts/vc-authn-oidc/templates/deployment.yaml#L73-L76

`/etc` directory is a system directory and contains configuration files for the operating system.
Placing web application files in the `/etc` directory introduces a security risk.

Relocating these files to `/app/config` would be more appropriate.

Gavinok:
This resolves #721 by moving code into /app/src and configuration into /app/controller-config


esune (approved):
👍🏻

PR #730, opened 2025-02-24
Use well known goal code for proof-request

esune:
Update goal code to use `aries.vc.verifier.once`, which indicates to the holder wallet the interaction is to be considered terminated after the first verification.


esune:
FYI @cvarjao @knguyenBC

loneil (approved):
Tried quickly locally, still works (added a dependabot fix in demo too)

PR #744, opened 2025-03-17
Support masking QR in darkmode browser extensions

### LINKED ISSUE #728: QR scan mask is offset a little (dark mode)
There's not a custom css or anything for dark mode on the included proof request frontend, noticed on Gavin's recent PR if the OS forces it to dark mode it's basically all fine (doesn't appear to really need any proper customization) but the "scan mask" overlay is a little off center in a way that's not noticeable in light mode.

Should just tweak the size by a few pixels so it looks more on-center.

Image

If a more customized color set for dark mode were ever needed, would leave that up to whoever wants their own custom templates.

Gavinok:
This resolves #728 by instead fading the image rather than trying to overlap divs which should be far more portable long term


esune (approved):
I like the approach 👍🏻

Gavinok:
@loneil Does this match what you are looking for?

Gavinok:
@loneil Does this match what you are looking for?

loneil (approved):
Looks good

PR #745, opened 2025-03-17
Update chart values, docs

esune:
Updated chart values file and documentation, removing old references and fixing current settings.

c.c.: @i5okie


charts/vc-authn-oidc/values.yaml
```diff
@@ -700,9 +696,8 @@ postgresql:
...
-        ## ref: https://github.com/bitnami/containers/tree/main/bitnami/postgresql#allow-settings-to-be-loaded-from-files-other-than-the-default-postgresqlconf
-        ##
-        primary:
-          extendedConfiguration: |
-            max_connections = 500
+    ## @param postgresql.primary.extendedConfiguration Extended PostgreSQL Primary configuration (appended to main or default configuration)
+    ## ref: https://github.com/bitnami/containers/tree/main/bitnami/postgresql#allow-settings-to-be-loaded-from-files-other-than-the-default-postgresqlconf
+    ##
+    extendedConfiguration: |
+      max_connections = 500
```
esune, line 703: I checked the bitnami repo and the old version seems wrong - this should fix things. See https://github.com/bitnami/charts/tree/main/bitnami/postgresql#postgresql-primary-parameters

charts/vc-authn-oidc/values.yaml
```diff
@@ -607,7 +604,7 @@ mongodb:
     ## If undefined (the default) or set to null, no storageClassName spec is
     ## set, choosing the default provisioner.
     ##
-    storageClass: netapp-block-standard
+    storageClass: default
```
esune, line 607: I believe this is correct, and a different storage provider can be specified if necessary by each deployer

i5okie (approved):

loneil (approved):
Assuming that `default` storage value is fine

PR #746, opened 2025-03-19
Basic browser detection and setting as query param in deeplink

### LINKED ISSUE #731: Same-device flow return hint
When completing a same-device authentication flow (i.e.: the user is not scanning a QR code on a separate device, but rather authenticating to a resource using a browser on the same device where the wallet is installed) the wallet has no easy way to know which browser application triggered the interaction via deep-link, and therefore it is impossible to direct the user automatically to the app (browser) initiating the flow.

It might be possible to detect the browser by processing the user-agent on the QR-code page, and add a query parameter to the deep-link URL so that the wallet could use it as a hint to determine which app to return to once the proof response has been submitted.

We need to determine:
- How reliable user-agent processing is to detect the specific app/browser initiating the auth flow
- How to embed this information in the deep link for consumption in the receiving app

Gavinok:
This is intended to resolve help resolve #731 
It simply appends a query param `browser` to the deeplink giving the name of the current browser. Here is an example where the name was `Mobile Safari`

```
bcwallet://aries_proof-request?c_i=eyJAaWQiOiAiZTN...aW9uIiwgInByaW9yaXR5IjogMH1dLCAiaGFuZHNoYWtlX3Byb3RvY29scyI6IFtdfQ==/?browser=Mobile%20Safari 
```

We can also integrate the new ua parser to detect the browser type as either mobile or desktop if we decide to go forward with this approach

Here are all the user options we can support matching against https://github.com/faisalman/ua-parser-js/blob/af8acf9078e5952a496d395464af8178ad55ccaf/src/enums/ua-parser-enums.js#L10-L163


loneil:
Note passing `browser=Mobile%20Safari` does inform the BC Wallet which deeplink prefix to use so somewhere would need to be a mapping for the actual app deeplink prefixes (for every single browser app?)

Also this wouldn't cover a case where a calling app (like the Google app) doesn't have it's own UA app identifier. IE on that what, from what I can tell, you'd just get "Chrome" and not know to deeplink back to the Google search app.

Maybe this limited support for major browsers would be fine...?

Gavinok:
https://github.com/bhagyas/app-urls and https://medium.com/@contact.jmeyers/complete-list-of-ios-url-schemes-for-third-party-apps-always-updated-5663ef15bdff seem to cover some of the possible schemas

esune:
> Note passing `browser=Mobile%20Safari` doesn't inform the BC Wallet which deeplink prefix to use so somewhere would need to be a mapping for the actual app deeplink prefixes (for every single browser app?)
> 
> Also this wouldn't cover a case where a calling app (like the Google app) doesn't have it's own UA app identifier. IE on that what, from what I can tell, you'd just get "Chrome" and not know to deeplink back to the Google search app.
> 
> Maybe this limited support for major browsers would be fine...?

I don't think the mapping should live in VC-AuthN since the browsers support is ultimately dependent on the app and which browsers are on the device.

html-templates/verified_credentials.html
```diff
@@ -285,11 +285,15 @@ <h5 v-if="state.showScanned" class="fw-bolder mb-3">
...
+  <script src="https://cdn.jsdelivr.net/npm/ua-parser-js/dist/ua-parser.min.js"></script>
   <script type="text/javascript">
     const { createApp, ref, reactive, computed, watch } = Vue;
-    /**
+   /**
      * @typedef  {"intro" |"verified" | "failed" | "pending"| "expired" |"abandoned"} AppStates
      */
+    // user agent detection
+    const uap = new UAParser();
+    console.log(uap.getResult())
```
esune, line 296: Nit.: if we leave a console.log statement (maybe as a `console.debug`) we may want to add some info about what this is - i.e.: "Detected user agent: browserUA"

html-templates/verified_credentials.html
```diff
@@ -287,11 +287,15 @@ <h5 v-if="state.showScanned" class="fw-bolder mb-3">
     </div>
   </script>
 
+  <script src="https://cdn.jsdelivr.net/npm/ua-parser-js/dist/ua-parser.min.js"></script>
```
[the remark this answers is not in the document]
Gavinok: removed the need for a cdn and just added the static js file. Figure it may just be easiest even if it's just for the bc wallet team to try this out

loneil:
Get an error trying the deep link on the wallet

image


Probably the encoded space `?browser=Mobile%20Safari`

but we likely shouldn't be passing that display name anyhow since it won't mean anything to the wallet, the wallet will have to know the custom URI for every browser it can go back to with this pattern anyways, so it would need to maintain some mapping (unless it's determined that URI mapping should be done in VCAuth but I don't think it's appropriate for the browser to be handling that). So if we go with this solution I would just use the key instead of the value of that mapping object (IE `SAFARI_MOBILE` for example)

Gavinok:
Ya as far as I know we are going to have to wait on this one to be supported by the wallet for it to work. I agree the mapping would likely be better just using the enum entry rather than the string. Due to the interdependence I still question if this is the best approach but don't have any better solution as of now

loneil:
Ok regardless of that we can't do the display name with a space in it I don't think, is there a way to access the key from that object instead of the value? "SAFARI_MOBILE" instead of "Mobile Safari", if I try that it doesn't blow up the wallet, it's just an unused parameter.

Gavinok:
Sure thing I'll change that now

html-templates/verified_credentials.html
```diff
@@ -242,7 +243,7 @@ <h5 v-if="state.showScanned" class="fw-bolder mb-3">
     <div>
       <a
         id="deep-link-button"
-        href="{{wallet_deep_link}}"
+        :href="combinedWalletLink"
```
loneil, line 246: There's a "DEBUG: Deep Link URL" part that's useful for checking the link, mind changing that to use this new field instead too?

loneil:
For thought later... since this is a specific pattern we are trying with BC implementation (might not be standard?) we could consider only putting this code in the BC gov template page and not the basic generic VCAtuth version, as someone can run VCAuth with a different deep link prefix that doesn't call the BC Wallet.

esune:
> For thought later... since this is a specific pattern we are trying with BC implementation (might not be standard?) we could consider only putting this code in the BC gov template page and not the basic generic VCAtuth version, as someone can run VCAuth with a different deep link prefix that doesn't call the BC Wallet.

That is a good point. The benefit of pushing it out in this repo is that we are more likely to get visibility/feedback from other users who may be facing (and resolving) the same problem already?

Gavinok:
I could not get the object from UAParser since it is not in the minified js, but I was able to replicate what is needed to get the browser. I removed the mobile info since, as of now, it seems unnecessary. I also added the info to the debugging element

html-templates/verified_credentials.html
```diff
@@ -289,9 +290,29 @@ <h5 v-if="state.showScanned" class="fw-bolder mb-3">
...
+   const getBrowserUA = () => {
+     // browsers
+     CHROME      = 'Chrome';
+     CHROMIUM    = 'Chromium';
+     CHROMECAST  = 'Chromecast';
+     EDGE        = 'Edge';
+     FIREFOX     = 'Firefox';
+     OPERA       = 'Opera';
+     FACEBOOK    = 'Facebook';
+     SOGOU       = 'Sogou';
```
esune, line 307: These don't seem to be used anywhere?

Gavinok:
found an edge case with Opera Mobile but it is fixed now. If any identifiers will contain a space we now simply go for the first word

esune (dismissed):
I think this works, we'll have to do more testing in conjunction with the wallet team to validate

loneil (changes requested):
Getting error on 1.0.23 called from iPad chrome right now, looking into but shouldn't merge at the momment

image

Gavinok:
@loneil Any idea what the user agent was?

html-templates/verified_credentials.html
```diff
@@ -289,9 +290,19 @@ <h5 v-if="state.showScanned" class="fw-bolder mb-3">
...
      * @typedef  {"intro" |"verified" | "failed" | "pending"| "expired" |"abandoned"} AppStates
      */
+    // user agent detection
+   const uap = new UAParser();
+   const getBrowserUA = () => {
+     PREFIX_MOBILE  = 'Mobile ';
+     // @type string
+     const b = uap.getResult().browser.name
+     const browserName = b.startsWith(PREFIX_MOBILE) ? b.split(" ")[1] : b.split(" ")[0]
+     return `{{wallet_deep_link}}?browser=${browserName.toUpperCase()}`;
```
esune, line 303: Possible cause of some of the issues: we should be appending `&browser=...` since the deep link payload is already a query parameter and the `?`defining the beginning of the query params namespace  is already present in the url

Gavinok: I'll give that a try. Good idea

html-templates/verified_credentials.html
```diff
@@ -289,9 +290,19 @@ <h5 v-if="state.showScanned" class="fw-bolder mb-3">
...
+    // user agent detection
+   const uap = new UAParser();
+   const getBrowserUA = () => {
+     PREFIX_MOBILE  = 'Mobile ';
+     // @type string
+     const b = uap.getResult().browser.name
+     const browserName = b.startsWith(PREFIX_MOBILE) ? b.split(" ")[1] : b.split(" ")[0]
+     return `{{wallet_deep_link}}&browser=${browserName.toUpperCase()}`;
+   }
+   console.log(`Detected Browser: ${uap.getResult().browser.name}`)
```
loneil, line 305: Should guard this code as well since we're calling a library (so anything that tries uap.getResult().browser.name) and doing string and array access so an exception could blow the render up.
Can try-catch around and log out an error to console, and handle so that then if the browser field is just not set the deep link will still work (as the wallet will have to handle missing/blank/unknown values in that anyways)

loneil:
> @loneil Any idea what the user agent was?

This happens on anything iOS, the URL format is invalid URL with the 2 `?`s in it

loneil (approved):
Think this looks good now. Opening in the BC Wallet it just sends the new unused `browser` query param around and it doesn't cause any errors. I forced a exception in the handling code and it worked as expected.

So think we can add this to the BC Gov template and deploy that to the dev env.

esune (approved):
👍🏻 let's do some extra testing once merged!

PR #752, opened 2025-04-03
Remove legacy configuration options

loneil:
`useOobPresentProof` - specified whether to use out of band, we can use that for everything now

`useUrlDeepLink` - was put in when we needed to feature toggle `?_url` on and off for testing

Clean up code that used these configuration values and remove them


oidc-controller/api/routers/oidc.py
```diff
@@ -143,39 +137,10 @@ async def get_authorize(request: Request, db: Database = Depends(get_db)):
 
     pres_exch_dict = response.model_dump()
 
-    # Prepeare the presentation request
+    # Prepear the presentation request
```
esune, line 140: Nit.: `prepear` 🙂

esune (commented):
Small typo, otherwise it's looking good.
Not really sure about why the code coverage would drop that significantly since tests were not removed - it all seems to be in ` /oidc-controller/api/core/config.py` in a utility function used to evaluate booleans.

loneil:
@esune I left in the USE_OOB_LOCAL_DID_SERVICE stuff, is that something we want to keep?

> Small typo, otherwise it's looking good. Not really sure about why the code coverage would drop that significantly since tests were not removed - it all seems to be in ` /oidc-controller/api/core/config.py` in a utility function used to evaluate booleans.

Yeah not sure why it counted a drop in coverage? Only 0.04% but I did not alter that code path... added a bit of coverage anyways, added a test for generating the proofs I noticed was lacking too.

esune (approved):

PR #765, opened 2025-05-12
Inform user of revoked credentials

### LINKED ISSUE #704: If a completed present proof 2.0 webhook is not verified (failed state) see what we can log and/or bring to the UI
In the event a proof is shared but does not pass actually pass the verification (we have this happen with unknown revocation issues with the BC Wallet interoperability lawyer cases before) the webhook comes back but is not verified, triggering our `failed` status. See some details and screenshot: https://github.com/openwallet-foundation/acapy-vc-authn-oidc/issues/703

In that code, if the webhook body DOES contain any relevant details about WHY it failed, we would like to return that to the UI as an expand error section. Kim has mockups.

Other/additonal options:

- Log out body or specific part on fail
- Adjust logging so we get an error level log?
- If webhook contains nothing useful, check if there's any ACA-Py enhancement that could be made to change that. Could check with community and raise an issue.

Gavinok:
This PR resolves #704 

I have added a new "Revoked" state to auth sessions that will be displayed as seen below when the provided credential has been revoked

image

In addition, the failed state has now had the revoked possibility removed from the list of possible errors. Something we should consider looking into is if the other possible errors would make sense. "Expired" credentials could mean a failed requested_predicate which we should specify. 

I also am not sure what we could really mean by a "missing" credential. Is this for when a wallet holds a credential which is not on the ledger? This should not be possible with the BC wallet


oidc-controller/api/routers/acapy_handler.py
```diff
@@ -54,9 +54,24 @@ async def post_topic(request: Request, topic: str, db: Database = Depends(get_db
                     if sid:
                         await sio.emit("status", {"status": "verified"}, to=sid)
                 else:
-                    auth_session.proof_status = AuthSessionState.FAILED
+                    client = AcapyClient()
+                    logger.info(
+                        f"MY RESULT {webhook_body['by_format']['pres']['indy']['identifiers'][0]['rev_reg_id']}"
```
esune, line 59: This is probably best as a `debug` log statement.

oidc-controller/api/routers/acapy_handler.py
```diff
@@ -54,9 +54,24 @@ async def post_topic(request: Request, topic: str, db: Database = Depends(get_db
...
+                    )
+                    rev = client.is_revoked(
+                        webhook_body["by_format"]["pres"]["indy"]["identifiers"][0][
+                            "rev_reg_id"
+                        ]
+                    )
+                    logger.info(f"Credential was revoked {rev}")
```
esune, line 66: This log statement might be best as a `warning`, so that it is highlighted in the logs if someone goes looking for information. Also wondering if there can be more details associated to the actual message, in order to determine which presentation exchange/credential triggered it.

esune (changes requested):
Thinking about the message in the UI: can we surface which credential is revoked, or just that "a credential" is revoked? In scenarios where the proof-request includes more than one credential restriction (e.g.: Person credential and LSBC credential for BCGov's lawyer use-case) it would be helpful to know which credential did not pass the revocation check. If this is not possible, we may just want to update the message to indicate that one or more crdentials that are part of the response to the proof-request ae revoked.

Gavinok:
> Thinking about the message in the UI: can we surface which credential is revoked, or just that "a credential" is revoked? In scenarios where the proof-request includes more than one credential restriction (e.g.: Person credential and LSBC credential for BCGov's lawyer use-case) it would be helpful to know which credential did not pass the revocation check. If this is not possible, we may just want to update the message to indicate that one or more crdentials that are part of the response to the proof-request ae revoked.

 Ya we can get the cred_def_id so from there we can get the tag. Would that be enough or is there something else we would need?

esune:
For `missing`: I do not think that use-case is currently possible, at least not with BC Wallet, as you would be prevented from submitting the response if you don;t hold all the credentials that are required. In general, it could be that the proof-request requires one or more credentials and the response does not send the correct ones in the response - unsure on how likely this scenario is though.



> > Thinking about the message in the UI: can we surface which credential is revoked, or just that "a credential" is revoked? In scenarios where the proof-request includes more than one credential restriction (e.g.: Person credential and LSBC credential for BCGov's lawyer use-case) it would be helpful to know which credential did not pass the revocation check. If this is not possible, we may just want to update the message to indicate that one or more crdentials that are part of the response to the proof-request ae revoked.
> 
> Ya we can get the cred_def_id so from there we can get the tag. Would that be enough or is there something else we would need?

For troubleshooting purposes knowing the `cred_def_id` would certainly be helpful as it would allow us to figure out which credential is causing the verification to fail.

Issue #740, opened 2025-03-14
Dependabot parsing problems


loneil:
I noticed there were some minor version updates in the pyproject/lock file that did not have dependabot PRs, on an existing PR I rebased I see this
`Dependabot can't parse your pyproject.toml. Because of this, Dependabot cannot update this pull request.`

Should confirm if there's an issue with dependabot after poetry upgrade?

loneil:
First step is probably to manually fire a dependabot check and see if it fails?

Gavinok:
Think I figured it out. The problem is coming from https://github.com/Gavinok/vc-authn-oidc/blob/main/.github/dependabot.yml#L16-L26 since it is only looking in /oidc-controller for dependancies it is missing the pyproject.toml all together. 

Using the dependabot cli seems to identify multiple dependancies when starting at the root 
~/go/bin/dependabot update pip openwallet-foundation/acapy-vc-authn-oidc

As opposed to identifying the specific directory which seems to provide no dependancy info at all
~/go/bin/dependabot update pip openwallet-foundation/acapy-vc-authn-oidc -d /oidc-controller

The solution would simply be to have it check the root of the project for pip dependancies

PR #774, opened 2025-06-12
Corrected dependabot path to pyproject.toml

Gavinok:
As explained in #740 

I think I've figured it out. The problem arises from https://github.com/Gavinok/vc-authn-oidc/blob/main/.github/dependabot.yml#L16-L26. Since it only looks in /oidc-controller for dependencies, it misses the pyproject.toml altogether.

Using the Dependabot CLI seems to identify multiple dependencies when starting at the root.
~/go/bin/dependabot update pip openwallet-foundation/acapy-vc-authn-oidc

In contrast, identifying a specific directory seems to provide no dependency info at all.
~/go/bin/dependabot update pip openwallet-foundation/acapy-vc-authn-oidc -d /oidc-controller

The solution is to have it check the root of the project for pip dependencies.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,7 +15,7 @@ updates:
   
   # Maintain dependencies for Python packages
   - package-ecosystem: "pip"
-    directory: "/oidc-controller"
+    directory: "/"
     schedule:
       interval: "weekly"
       day: "monday"


.github/dependabot.yml
```diff
@@ -15,7 +15,7 @@ updates:
   
   # Maintain dependencies for Python packages
   - package-ecosystem: "pip"
-    directory: "/oidc-controller"
+    directory: "/"
```
loneil, line 18: Good catch, that must be it. 
Do you think there's an ignore that would be worth adding here (if that's even possible for this dependabot yml syntax), as the checks below it go into `demo` and `docker` so might be worth having this one avoid those folders that are already scanned with a different ecosystem? I know it's specifying pip so probably won't find any python in there anyways...
Could be overkill unless it's an easy pattern

loneil (approved):

PR #801, opened 2025-07-16
build(vue.js): Update Keycloak proof id and configuration

Gavinok:
- Removes hardcoded configuration for the proof presentation request id
- Integrates environment variable FRONTEND_KC_PRES_REQ_CONF_ID
- Updates docker-compose.yaml to use 'showcase-person' instead of 'test-proof' collection in FRONTEND_KC_PRES_REQ_CONF_ID variable


demo/vue/app/frontend/src/store/modules/auth.js
```diff
@@ -77,7 +76,7 @@ export default {
         window.location.replace(
           getters.createLoginUrl(options) +
             '&pres_req_conf_id=' +
-            getters.presReqConfId
+            getters.presReqConfId,
```
esune, line 79: I can't find this getter in the current file - am I missing something?

loneil (approved):
Looks good, uses var from docker-compose locally

image

demo/vue/app/frontend/src/store/modules/auth.js
esune: I really did not see it 😅

PR #811, opened 2025-07-26
Jul 25 Dependency updates

loneil:
Regular library updates


loneil:
Not sure why coveralls complaining about a (negligable) decrease in coverage. No changes other than dependencies so it shouldn't be scanning lock files or anything.

PR #813, opened 2025-08-05
Implement socket ID management in AuthSession for High Availability

### LINKED ISSUE #812: Implement MongoDB-based WebSocket ID tracking for High Availability
Following the spike investigation in #805, we need to implement the MongoDB-based solution for maintaining WebSocket connection state to support high-availability deployments.

Current Problem:
- WebSocket IDs are stored in-memory dictionary
- Service replicas scaling down severs WebSocket connections
- No way to restore connection context after reconnection

Proposed Solution:
Replace the in-memory WebSocket ID storage with MongoDB persistence to maintain transaction context during reconnections.

Acceptance Criteria:
- Remove in-memory dictionary for WebSocket ID storage
- Implement MongoDB collection for tracking WebSocket connection states
- Ensure latest transaction state can be read from database on client reconnect
- Implement WebSocket reconnection handling that restores previous state
- Add appropriate cleanup for stale WebSocket records
- Test scaling scenarios (scale up/down) to verify connections remain functional
- Verify no data loss during service replica changes

Technical Notes:
- This builds on the research from the HA spike (#805)
- Should maintain backward compatibility during transition
- Consider WebSocket ID lifecycle management and cleanup strategies

Gavinok:
This PR resolves #812 by replacing the in memory websocket connection managment with a MongoDB based solution. 

To do this I added websocket_id to the AuthSession and utilize it as follows:

- Introduced `get_by_socket_id` method in `AuthSessionCRUD` for retrieval by socket ID.
- Modified `initialize` event in `socketio` to store the socket ID in the corresponding `AuthSession`.
- Updated `disconnect` event to clear the socket ID from the `AuthSession` when a client disconnects.
- Replaced direct connections management with asynchronous methods to obtain socket IDs for presentation IDs in relevant router functions.

### DIFF
--- a/oidc-controller/api/authSessions/models.py
+++ b/oidc-controller/api/authSessions/models.py
@@ -27,6 +27,7 @@ class AuthSessionBase(BaseModel):
     pyop_auth_code: str
     response_url: str
     presentation_request_msg: dict | None = None
+    socket_id: str | None = None
     model_config = ConfigDict(populate_by_name=True)
     created_at: datetime = Field(default_factory=datetime.utcnow)
 
--- a/oidc-controller/api/authSessions/crud.py
+++ b/oidc-controller/api/authSessions/crud.py
@@ -89,3 +89,12 @@ class AuthSessionCRUD:
             )
 
         return AuthSession(**auth_sess)
+
+    async def get_by_socket_id(self, socket_id: str) -> AuthSession | None:
+        col = self._db.get_collection(COLLECTION_NAMES.AUTH_SESSION)
+        auth_sess = col.find_one({"socket_id": socket_id})
+
+        if auth_sess is None:
+            return None
+
+        return AuthSession(**auth_sess)
--- a/oidc-controller/api/routers/acapy_handler.py
+++ b/oidc-controller/api/routers/acapy_handler.py
@@ -11,7 +11,7 @@ from ..authSessions.models import AuthSession, AuthSessionPatch, AuthSessionStat
 from ..db.session import get_db
 
 from ..core.config import settings
-from ..routers.socketio import sio, connections_reload
+from ..routers.socketio import sio, get_socket_id_for_pid
 
 logger: structlog.typing.FilteringBoundLogger = structlog.getLogger(__name__)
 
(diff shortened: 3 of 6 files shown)


esune (commented):
Changes look good to me, will see if I can run and test it locally with multiple instances otherwise we may have to test on a deployed environment.

oidc-controller/api/authSessions/models.py
```diff
@@ -27,6 +27,7 @@ class AuthSessionBase(BaseModel):
     pyop_auth_code: str
     response_url: str
     presentation_request_msg: dict | None = None
+    socket_id: str | None = None
```
loneil, line 30: Think we should index this field since there's a `get by` I think, and be good to add a unique constraint? 

I don't think we'd be adding an SID to more than one auth session record right? Or I could be wrong about that and we create new auth sessions that keep the SID... if not I'd do a unique constraint (would notice pretty quick if it breaks)

loneil: Yeah trying out locally think we can go with a unique constraint on the SID + the index

Gavinok:
Hold up found an error

Gavinok:
@loneil ok should be good to go.

esune (dismissed):
LGTM, letting @loneil merge if he is happy with the changes

oidc-controller/api/db/session.py
```diff
@@ -94,6 +94,12 @@ async def init_db():
     auth_session = db.get_collection(COLLECTION_NAMES.AUTH_SESSION)
     auth_session.create_index([("pres_exch_id", ASCENDING)], unique=True)
     auth_session.create_index([("pyop_auth_code", ASCENDING)], unique=True)
+    # Partial unique index for socket_id - only indexes non-null values, ensuring uniqueness
+    auth_session.create_index(
+        [("socket_id", ASCENDING)],
+        unique=True,
+        partialFilterExpression={"socket_id": {"$exists": True, "$type": "string"}},
```
loneil, line 101: Notice when putting profiling on in Mongo and doing a socket lookup (ie, when we do `auth_sess = col.find_one({"socket_id": socket_id})` in the code above) that it ends up doing a COLLSCAN

image

So not using the index...

This index includes documents where socket_id exists and is a string. However, MongoDB’s query planner does not always match this partial filter expression with a query on read that way.

Looking it up there seems to be a way people handle this
`partialFilterExpression={"socket_id": { "$gt": "" }}`
which is a little weird (greater than?)

But works and does an index seek

image

Gavinok: Ya not what I would expect. I'll give it a try

loneil (approved):
Looking good, can try some additional pods on dev. Merging

PR #828, opened 2025-08-25
Make all created and expired timestamps rather than strings

### LINKED ISSUE #824: A new auth_session writes created_at and expired_timestamp as strings then they update as dates
See `created_at` and `expired_timestamp` values being string and datetime differently below.
Should check on verify, expired, abandoned, etc states as well for this ticket.

Image

Gavinok:
This PR resolves #824 by inserting auth_sessions using a model dump
rather than JSON encoding them. This prevents dates being encoded as a
string rather than a timestamp.


loneil (approved):
Looks good to me

PR #834, opened 2025-08-26
Manually handle deletion of exchange records

### LINKED ISSUE #823: Handle present proof 2.0 outside of webhook
as of https://github.com/openwallet-foundation/acapy/issues/3803 we can nolonger perform present proof 2.0. For more details see https://github.com/openwallet-foundation/acapy/issues/3065

Gavinok:
This PR resolves #823 by using `--preserve-exchange-records` and retrieving the presentation_request manually then deleting the record.

To take advantage of the connection deletion we recently integrated I have combined the connection and 

On the off chance that a record has not been deleted e.g. client pod
crashed, exchange was never verified, etc it will be deleted by a
dedicated `cleanup_service` running in the background. 

The service is configured with
`CONTROLLER_PRESENTATION_RECORD_RETENTION_HOURS` and
`CONTROLLER_PRESENTATION_CLEANUP_SCHEDULE_MINUTES` which determines
how often the service is ran and how old the records need to be before
they are deleted.

## Possible changes

- Presentation data retrieval failure handling could handle race conditions better
- Consider pagination for large record sets in production


jamshale (commented):
Looking good 👍  I agree with Emiliano's comments. I'm glad we are doing this now. I feel like it would definitely get caught on a security audit.

oidc-controller/api/core/acapy/client.py
```diff
@@ -244,14 +294,88 @@ def delete_connection(self, connection_id: str) -> bool:
...
+                        f"Failed to delete presentation record {presentation_exchange_id}"
+                    )
+
+            except Exception as e:
+                error_msg = f"Error deleting presentation record {presentation_exchange_id}: {e}"
+                errors.append(error_msg)
+                logger.error(error_msg)
+
+        # Second, delete the connection if provided
+        # TODO: make manditory when connections we drop OOB
```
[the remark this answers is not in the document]
Gavinok: Think I changed my workding and forgot to look it over

oidc-controller/api/services/cleanup.py
```diff
@@ -0,0 +1,137 @@
...
+                        logger.warning(
+                            f"Failed to parse timestamp {created_at_str} for record {record.get('pres_ex_id')}: {parse_error}"
+                        )
+                        continue
+
+                    # Check if record is old enough to clean up
+                    if record_time < cutoff_time:
+                        pres_ex_id = record.get("pres_ex_id")
+                        presentation_deleted, _, errors = (
+                            self.client.delete_presentation_record_and_connection(
```
loneil, line 67: O(n) API calls over to ACA-Py is... going to be ok for heavy load?

jamshale: Haha. Might be a bit of an experiment. The delete endpoint is very simple look up and delete by indexed key so it should be performant.

Gavinok:
Good catch @esune I have fixed that in #825 and will merge it here

esune:
I tested this, works fine for presentation exchanges that complete (i.e.: a credential is shared) and verify, however `abandoned` presentation exchanges do not get deleted once the flow is complete. Is the intention to NOT delete those immediately and leave them for the cleanup process to figure out? Can we handle deletion by acting on the problem-report that is sent back?

Nit.: to test the retention policy it might be worth indicating the retention time in minutes instead, so we can set it to a low number for local development purposes (current minimum is 1hr).

oidc-controller/api/services/cleanup.py
```diff
@@ -0,0 +1,143 @@
...
+
+                # Sleep for the configured interval
+                sleep_seconds = self.schedule_minutes * 60
+                logger.debug(f"Background cleanup sleeping for {sleep_seconds} seconds")
+                await asyncio.sleep(sleep_seconds)
+
+            except Exception as e:
+                logger.error(f"Background cleanup task encountered an error: {e}")
+                # Sleep for a shorter period on error to retry sooner
+                await asyncio.sleep(300)  # 5 minutes on error
```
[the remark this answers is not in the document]
esune: @Gavinok does the refactoring using `apscheduler` update this?

oidc-controller/api/main.py
```diff
@@ -132,13 +134,22 @@ async def on_tenant_startup():
     """Register any events we need to respond to."""
     await init_db()
     await init_provider(await get_db())
+
+    # Start the APScheduler-based cleanup service
+    await cleanup_service.start_scheduler()
+    logger.info("Started APScheduler-based presentation record cleanup service")
```
esune, line 140: This depends on Redis, we should NOT attempt to start this up if Redis is not being used.
Probably best to handle this inside the `start_scheduler` function so we can log success/failure depending on a self-check.

esune (changes requested):
Small tweak to make sure we don;t have issues when starting the service if Redis is not available.

esune (approved):
This looks good, thank you for all the work @Gavinok !
Let's get this merged and add the cron job to the helm chart now 🙂

PR #855, opened 2025-09-10
[chart] Update Acapy sub-chart to version 0.2.1 to use new secrets values

i5okie:
The improved Acapy helm chart allows specifying existing secrets. This is required for a smooth upgrade of test and production environments.


loneil (dismissed):
give a try in dev

Gavinok (dismissed):

PR #926, opened 2025-11-29
Feat: add traction tenant mode for secure webhook registration

### LINKED ISSUE #923: Feat: support Traction tenant API fallback for webhook registration
**Problem:**
When deploying the OIDC Controller against a Traction instance (or any secured ACA-Py multi-tenant deployment), the `/multitenancy/wallet/{wallet_id}` Admin API endpoints could be blocked or restricted (returning 403 Forbidden or 404 Not Found) for security reasons by an adminstrator.

Currently, the controller attempts to register its webhook URL exclusively via this Admin API endpoint on startup. If this fails, the webhook is never registered, and the authentication flow hangs.

**Root Cause Analysis:**
The current `register_tenant_webhook` function assumes access to the privileged Admin API which typically requires the global `admin-api-key`. Traction deployments often block these paths (`/multitenancy/wallet/...`) to prevent exposure of the super-admin key.

Instead of reliance on endpoints that require the admin API key, Traction offers alternative tenant-centric endpoints that can be used to accomplish the same configuration changes using a tenant-specific Bearer token.

Specifically, the `/tenant/wallet` endpoint allows updating the wallet configuration (including webhook URLs) when authenticated with a token derived from the wallet key.

**Expected Behavior:**
If the Admin API call fails with a 403 or 404, the Controller should attempt to:
1.  Authenticate as the specific tenant using the configured Wallet Key (`MT_ACAPY_WALLET_KEY`) to obtain a token (via `/multitenancy/wallet/{wallet_id}/token`).
2.  Use the `/tenant/wallet` endpoint with the Bearer token to update the webhook URL.

**Suggested Solution:**
Refactor `register_tenant_webhook` to accept a `token_fetcher` callable. If the primary Admin API call fails:
1.  Invoke the fetcher to get a Bearer token.
2.  Call `PUT /tenant/wallet` with the new webhook configuration.

MonolithicMonk:
**Reference:** Closes #923

**Summary:**
This PR introduces full support for integrating with Traction (or other secured multi-tenant ACA-Py deployments) where the privileged Admin API endpoints are restricted.

*   **New Tenancy Mode:** Introduces `AGENT_TENANT_MODE="traction"` (mapped to `ACAPY_TENANCY`).
*   **Authentication:** Adds `TractionTenantAcapy` configuration class to handle authentication via `TRACTION_TENANT_ID` and `TRACTION_TENANT_API_KEY` (preferred), with a fallback to Wallet ID/Key authentication.
*   **Secure Webhook Registration:** Refactors `register_tenant_webhook` to support a "direct" mode. When in Traction mode, the controller bypasses the restricted Admin API (`PUT /multitenancy/wallet/{id}`) and uses the authenticated Tenant API (`PUT /tenant/wallet`) directly.
*   **Configuration:**
    *   Updates `docker-compose` and `manage` scripts to pass through `TRACTION_` variables.
*   **Documentation:** Updates `ConfigurationGuide.md` and `MigrationGuide.md` with new variables and startup requirements.
*   **Testing:** Adds comprehensive unit tests for the new configuration logic, client initialization, and webhook registration fallback strategies.

**Impact:**
Enables VC-AuthN to be deployed in hardened environments where the ACA-Py Admin API is blocked for security, preventing the "webhook registration hanging" issue described in #923.

### DIFF
--- a/oidc-controller/api/core/config.py
+++ b/oidc-controller/api/core/config.py
@@ -332,3 +332,9 @@ if (
         "The controller will not be able to register webhooks with the tenant wallet, "
         "which may cause verification flows to hang."
     )
+
+# Startup validation for ACAPY_TOKEN_CACHE_TTL
+if settings.ACAPY_TOKEN_CACHE_TTL <= 0:
+    raise ValueError(
+        f"ACAPY_TOKEN_CACHE_TTL must be a positive integer, got '{settings.ACAPY_TOKEN_CACHE_TTL}'"
+    )
--- a/oidc-controller/api/core/tests/test_config_coverage.py
+++ b/oidc-controller/api/core/tests/test_config_coverage.py
@@ -94,3 +94,17 @@ class TestConfigEdgeCases:
 
                 for call in mock_logger.warning.call_args_list:
                     assert "ACAPY_TENANCY is set to 'multi'" not in call[0][0]
+
+    def test_token_ttl_validation_failure(self):
+        """
+        Test that ValueError is raised when ACAPY_TOKEN_CACHE_TTL is invalid (<= 0).
+        """
+        with patch.dict(os.environ, {"ACAPY_TOKEN_CACHE_TTL": "0"}):
+            with pytest.raises(ValueError) as exc:
+                self.reload_config()
+            assert "ACAPY_TOKEN_CACHE_TTL must be a positive integer" in str(exc.value)
+
+        with patch.dict(os.environ, {"ACAPY_TOKEN_CACHE_TTL": "-10"}):
+            with pytest.raises(ValueError) as exc:
+                self.reload_config()
+            assert "ACAPY_TOKEN_CACHE_TTL must be a positive integer" in str(exc.value)


docker/.env.example
```diff
@@ -0,0 +1,132 @@
...
+# Traction / ACA-Py admin endpoints
+AGENT_ADMIN_URL=https://traction-admin.example.com
+AGENT_ENDPOINT=https://traction-acapy-endpoint.example.com
+
+AGENT_ADMIN_API_KEY=changeme
+AGENT_GENESIS_URL=https://test.bcovrin.vonx.io/genesis
+AGENT_WALLET_SEED=your-32-char-seed-here-00000000000000
+
+
+############################################
```
loneil: I like this, allowing both types of access in a single setting 👍 
Traction could still use wallet id/key, though API key is preferable.

docs/ConfigurationGuide.md
```diff
@@ -85,15 +88,57 @@ Several functions in ACAPy VC-AuthN can be tweaked by using the following enviro
...
 | LOG_LEVEL                 | "DEBUG", "INFO", "WARNING", or "ERROR" | sets the minimum log level that will be printed to standard out                                                                                                                                                                                                                                                                                                                                                                                        | Defaults to DEBUG                                                                                                                       |
 
-### Multi-Tenant Webhook Registration
+### Legacy Variable Support
```
loneil: Understand keeping these around for backwards compatibility, but am understanding right that the Multitenancy didn't really work until @MonolithicMonk 's recent fixes? In that case, or in any case, what are people's thoughts on deprecating them? Maybe not right away but eventually, just to avoid config drift? Especially if it's unlikely they were ever used in the wild (unless I'm mistaken about that)

Possibly nothing to do here (unless we know for sure they could be deleted) but could raise an issue if we think that's a good idea.

oidc-controller/api/core/acapy/config.py
```diff
@@ -15,12 +15,15 @@ def get_headers() -> dict[str, str]: ...
 
 
 class MultiTenantAcapy:
```
loneil: It occurs to me that, while not an issue with Traction, could this multi tenant token fetch be a problem on standard ACA-Py if a VCAuth instance is running with multiple processes (VCAuth can start up locally like that to try it out). A reason we use the [multitenant provider plugin](https://github.com/openwallet-foundation/acapy-plugins/tree/main/multitenant_provider) is that we weren't able to get more than 1 token at once. So if I start up 3 pods in a scalable deployment, would 2 of the pods on a regular ACA-Py possibly have trouble as the issuance of the other tokens invalidate the existing ones?

oidc-controller/api/core/acapy/config.py
```diff
@@ -65,6 +68,62 @@ def get_headers(self) -> dict[str, str]:
         return {"Authorization": "Bearer " + self.get_wallet_token()}
 
 
+class TractionTenantAcapy:
```
loneil: In Traction the multitenant tokens expire with a default of 1 day configurable. VCAuth processes/containers would be running past that (or the expiry could be very short if desired and still be shorter than the life of a serverless process runtime) would token failures trigger a re-fetch?

loneil (changes requested):
This is really great to get going @MonolithicMonk . I tried this PR out locally with my vcauth using the hosted [Traction Sandbox](https://traction-sandbox-tenant-ui.apps.silver.devops.gov.bc.ca/) and it was working for me. Great to be able to watch connection and verification go through on the Tenant UI as testing.

I just have a concern about token expiry to look at I think.

docker/.env.example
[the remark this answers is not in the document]
loneil: Yeah I never liked that defaults can kind of be spread out like that, agree that setting all these envs in an actual env file (and it's nice that this PR adds the sample one) seems better than in the manage script.

oidc-controller/api/core/acapy/config.py
```diff
@@ -56,10 +69,83 @@ def get_wallet_token(self):
...
+    """
+
+    # Map unified variables to Traction concepts
+    tenant_id = settings.ACAPY_TENANT_WALLET_ID
+    api_key = settings.ACAPY_TENANT_WALLET_KEY
+
+    # Class-level cache
+    _token: str | None = None
+    _token_expiry: float = 0.0
+    TOKEN_TTL: int = 3600
```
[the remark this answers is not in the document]
esune: +1 to this, better to keep it conigurable

PR #930, opened 2025-12-04
Hotfix/v2.3.4

esune:
Fixing an edge case where the id_token still didn't include the VC claims if the subject identifier was not provided or consistently generated.


jamshale (approved):
Not my area of expertize, but the comments make sense.

PR #971, opened 2026-02-23
Replace requests with httpx for async HTTP                                                                                                                                                       

### LINKED ISSUE #863: Async HTTP Requests
Currently we make heavy use of the requests library. While good for syncronous work we should eventually move to httpx to allow for async requests.

Gavinok:
This PR resolves #863
# Why this was needed
                                                                                                
The controller is a FastAPI async application, but all HTTP calls to ACA-Py were made using the requests library, which is synchronous. Every call to ACA-Py (create presentation request, fetch wallet DID, delete records, register webhooks) was blocking the event loop for the duration of the network round trip. Under load with multiple replicas this degrades throughput across the board since no other coroutine can run while a request is in flight.

# What changed

## Single shared httpx.AsyncClient
Instead of creating a new HTTP connection for every request, a single httpx.AsyncClient is created once at application startup and stored on app.state.http_client. It is closed cleanly on shutdown. This means connections to ACA-Py are kept alive and reused across requests (keepalive pool of 20 connections, max 100 total), which cuts connection overhead significantly compared to the old approach of opening a new TCP connection on every call.

## All ACA-Py calls are now properly async
Every method on AcapyClient that previously called requests.get/post/delete is now async def and uses await self._http_client.get/post/delete. The client instance is injected at construction time (AcapyClient(http_client)) rather than created internally, making it easy to pass the shared client from the router layer and easy to mock in tests.

## Token refresh lock (MultiTenantAcapy, TractionTenantAcapy)
In multi-tenant and traction modes the controller needs a short-lived bearer token to talk to ACA-Py. With async, many coroutines can be waiting for a token at the same time. Without coordination they would all see an expired token simultaneously and all fire off a token refresh request at once, which wastes connections and can hit rate limits on the ACA-Py admin API.

The fix is a class-level asyncio.Lock with a double-checked locking pattern:

1. Check the cache without the lock (fast path - no overhead for the common case)
2. If the token is expired, acquire the lock
3. Check the cache again inside the lock (another coroutine may have already refreshed it while
we were waiting)
4. If still expired, fetch a new token, cache it, release the lock

The result is that no matter how many concurrent requests come in with an expired token, only one HTTP call goes out to refresh it. All other waiters reuse the result.

The lock is a class-level attribute so it is shared across all instances that use the same tenant credentials. It is replaced with a fresh asyncio.Lock() in test fixtures so a failed test that holds the lock cannot deadlock the next test.

## webhook_utils updated

register_tenant_webhook and _register_via_tenant_api now accept and use httpx.AsyncClient instead of calling requests directly. The token_fetcher type hint was updated from Callable[[], str] to Callable[[], Awaitable[str]] to reflect that it is now an async callable.

# Things to consider going forward

Timeout handling is inconsistent. Methods with try/except (delete_connection, delete_presentation_record, send_problem_report) catch all exceptions including timeouts and return a safe default. Methods without try/except (create_presentation_request, get_wallet_did, create_connection_invitation) let httpx.ReadTimeout propagate to the caller. The routers do not currently catch this, so a timeout on a proof request creation will result in a 500 to the relying party. A follow-up should either add a consistent timeout handling strategy or wrap those calls in the router layer.

The `httpx.AsyncClient` timeout is set globally at 10 seconds. This is fine for most calls but proof request creation and connection management calls can be slower if ACA-Py is under load. Consider making the timeout configurable via an environment variable.

`get_connections_batched` silently stops on error. If a network error occurs mid-pagination the generator stops and returns what it has fetched so far. This means a cleanup run could silently process fewer connections than expected. This is the safe choice for now but it should be logged more visibly so operators can detect partial runs.

### DIFF
--- a/docker/manage
+++ b/docker/manage
@@ -281,13 +281,14 @@ sedInPlace() {
 initializeWalletSeed() {
   # Generate a new wallet seed and update it in .env
   _newSeed=$(generateSeed vc-authn-oidc)
-  sedInPlace "s/^AGENT_WALLET_SEED=.*/AGENT_WALLET_SEED=${_newSeed}/" .env
+  # Use | as sed delimiter — base64 seeds may contain / which breaks s///
+  sedInPlace "s|^AGENT_WALLET_SEED=.*|AGENT_WALLET_SEED=${_newSeed}|" .env
   echoSuccess "Generated new AGENT_WALLET_SEED in .env"
 }
 
 resetWalletSeed() {
   # Reset wallet seed to default value in .env
-  sedInPlace "s/^AGENT_WALLET_SEED=.*/AGENT_WALLET_SEED=${DEFAULT_AGENT_WALLET_SEED}/" .env
+  sedInPlace "s|^AGENT_WALLET_SEED=.*|AGENT_WALLET_SEED=${DEFAULT_AGENT_WALLET_SEED}|" .env
   echoInfo "Reset AGENT_WALLET_SEED to default value in .env"
 }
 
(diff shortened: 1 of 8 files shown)


oidc-controller/api/core/acapy/client.py
```diff
@@ -383,84 +315,40 @@ def get_connections_batched(self, state: str = "invitation", batch_size: int = 1
             f"<<< get_connections_batched -> yielded {total_yielded} total connections"
         )
 
-    def get_all_connections(self) -> list[dict]:
```
esune: Was `get_all_connections` unused?

Gavinok: Ya it's not accessed in the code base

PR #980, opened 2026-03-03
No Longer Preserve Records and Rely On Debug-Webhooks

### LINKED ISSUE #972: Use "debug" webhooks with aut-delete flag for presentations
In #834 we introduced code to manually delete exchange records after the removal of full webhook payload from ACA-Py. After using this for a few months, we realized that this is unnecessarily complicating the code and we should revert to the previous logic by using the `debug` webhooks flag in ACA-Py. From a security standpoint this does not cause issues in a normal deployment, since the agent and controller talk to each other over the internal, non-exposed network.

The options we have are:
1. Fully revert changes, removing all the related code
2. Revert changes so that exchange records are deleted automatically, clean-up manual handlers, but leave the cleanup endpoints intact for purging if/when needed

While I think #2 is a legitimate use-case, I think that it is equally easy, as an administrator, to get access to the db and clean things up as necessary.

@loneil @Gavinok please chime in with your perspective, how do you think we should move forward?

Gavinok:
This PR resolves #972

### DIFF
--- a/oidc-controller/api/routers/tests/test_acapy_handler.py
+++ b/oidc-controller/api/routers/tests/test_acapy_handler.py
@@ -338,9 +338,6 @@ class TestConnectionBasedVerificationWebhooks:
             "status", {"status": "abandoned"}, to="test-socket-id"
         )
 
-    @pytest.mark.skip(
-        reason="Expiration logic in handler has implementation issue - test skipped for now"
-    )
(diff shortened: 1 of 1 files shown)


esune (changes requested):
A couple tiny tweaks and some tests failures, otherwise looking good

PR #1035, opened 2026-04-20
Adjust main.py tzset call for windows to fix pytest

### LINKED ISSUE #1036: Poetry run pytest breaks on Windows (AttributeError: module 'time' has no attribute 'tzset')
Can't run `poetry run pytest` on Windows any more as some newer integration tests invoke `main.py` running through `tzset()` which is Unix only.

```
ERROR oidc-controller/api/integration_tests - AttributeError: module 'time' has no attribute 'tzset'
ERROR oidc-controller/api/tests/test_main_integration.py - AttributeError: module 'time' has no attribute 'tzset'
ERROR oidc-controller/api/tests/test_webhook_registration.py - AttributeError: module 'time' has no attribute 'tzset'
```

This is just on the app startup so it's not that these tests broke anything, but just that the app would probably never have started up directly on Windows at all (this is old code). VCAuthN is always run through the Docker containers anyways so can just handle this by skipping it for Win.

loneil:
Addresses https://github.com/openwallet-foundation/acapy-vc-authn-oidc/issues/1036

Guard the Unix-only time.tzset() call in `main.py` so module import no longer crashes on Windows. Unblocks pytest collection for contributors developing on Windows.

The change is behaviorally identical on Linux, which is the only supported runtime (IE the Docker images we publish).

On Linux/macOS, `getattr` returns the real time.tzset and it's invoked exactly as before. So same for prod, CI, and Docker. The only diff is on Windows, where `time.tzset` doesn't exist in CPython at all; there, we now skip the call.

"This is the idiomatic pattern for Unix-only stdlib APIs and is arguably a latent portability fix, not just a test-env convenience: the previous code would have always failed on Windows, it just hadn't been exercised."

### DIFF
--- a/oidc-controller/api/main.py
+++ b/oidc-controller/api/main.py
@@ -55,7 +55,9 @@ logging_file_path = os.environ.get(
 
 
 os.environ["TZ"] = settings.TIMEZONE
-time.tzset()
+tzset = getattr(time, "tzset", None)
+if tzset is not None:
+    tzset()
 
 
 def get_application() -> FastAPI:


jamshale (approved):

oidc-controller/api/main.py
```diff
@@ -55,7 +55,9 @@
 
 
 os.environ["TZ"] = settings.TIMEZONE
-time.tzset()
+tzset = getattr(time, "tzset", None)
+if tzset is not None:
+    tzset()
```
jamshale, line 60: Could also do 
```
tzset = getattr(time, "tzset", None)
if callable(tzset):
    tzset()
```
but what you did is fine.

esune (approved):

PR #1044, opened 2026-05-05
Reconnection for SSEs

loneil:
When a user returns to the verification page after a wallet deep-link backgrounded the tab, force an immediate `EventSource` reconnect instead of waiting on the browser's native backoff.

The page now wraps `EventSource` setup in a connect() helper and listens for `visibilitychange`: when the tab becomes visible again and a connection is still open, it closes and reopens the stream. On each new connection the server's _sse_event_loop re-reads proof status from the DB and emits it as the initial event, so reconnecting deterministically resyncs state — including any terminal status (**verified / failed / expired / abandoned**) that fired while the tab was hidden and was missed by the suspended EventSource. Once a terminal status is received, the client closes the stream and stops reconnecting.

Adds a log line on terminal initial-state emits (with a `last-event-id` reconnect flag) to confirm the recovery path in prod, plus tests for the SSE router.

### DIFF
--- a/html-templates/verified_credentials.html
+++ b/html-templates/verified_credentials.html
@@ -375,23 +375,39 @@
         },
       },
       mounted() {
-        /**
-         * Initialize SSE connection for real-time status updates.
-         * EventSource auto-reconnects natively on disconnect.
-         */
-        const evtSource = new EventSource('/sse/status/{{pid}}');
-        evtSource.addEventListener('status', (e) => {
-          const { status } = JSON.parse(e.data);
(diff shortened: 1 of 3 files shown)


esune (dismissed):

Gavinok (dismissed):

esune:
@loneil it looks like one of the new unit tests might be timing out and never failing, leading to the testing step to timeout without completing. If you have a minute to check it would be good, otherwise I can take a peek.

esune (approved):

PR #1101, opened 2026-08-11
Update ACA-Py to 1.6.1

loneil:
Update aca-py image to 1.6.1


demo/nuxt/package.json
```diff
@@ -10,7 +10,11 @@
   },
   "dependencies": {
     "bootstrap": "^5.3.3",
-    "nuxt": "^4.4.7",
+    "nuxt": "^4.5.2",
     "nuxt-oidc-auth": "^1.0.0-beta.10"
+  },
```
loneil, line 15: Pin esbuild ≥0.28.2 via overrides to remediate the esbuild dev-server CORS advisory (GHSA-67mh-4wv8-2f99) pulled in transitively; bump @nuxt/devtools-ui-kit to a version compatible with the patched esbuild. Temporary until upstream dependency ranges catch up.

Dependabot flagged that advisory as **critical**, so addressing even though nuxt is slow to get it

Gavinok (approved):

